PR #2, opened 2017-10-02
Travis: Use current user inside docker

fwalch:


ch1bo (approved):

fwalch:
This doesn't work because ROS needs a home directory to write some logs and the user given with `-u` doesn't have one inside the container. As we don't need to use data generated inside the container from outside (unlike in the libfranka build), I'll just close this for now.

PR #6, opened 2017-10-08
hw: add missing test_depend on description pkg.

gavanderhoorn:
As per subject.

Looking through the files in `franka_hw`, I believe `franka_description` is only a `test_depend`, so I added it as such to the manifest.


gavanderhoorn:
I'm building `franka_ros` pkgs on a highly parallel machine (72 jobs) so perhaps that is what triggered this.

fwalch (approved):

fwalch:
Thanks!

PR #11, opened 2018-02-06
Adding missing dependency to package.xml

mlautman:


fwalch:
Thanks for the PR, but shouldn't this be an `exec_depend`? Also, can you change the target branch to `develop` instead of `kinetic-devel`?

mlautman:
done

PR #15, opened 2018-04-22
Update control_mode.cpp

User-TGK:
Package franka_hw doesn't build without the vector include statement.


fwalch:
Thanks for the contribution! It looks like the file you added the `#include` to doesn't need it (no use of `std::vector`). Can you add it to the correct file?

It does look like `control_mode.h` is missing an include of `<type_traits>` (for `std::underlying_type_t`), though.

fwalch:
Thanks!

PR #52, opened 2019-03-03
Panda in Gazebo + MoveIt! motion planning

erdalpekel:
Dear team franka_ros,

I have integrated the panda robot in gazebo and would like to contribute this to the repository so that everybody can use it.
I have created this pull request based on my [blog post](https://erdalpekel.de/?p=55).

There is a known issue though: I didn't want to modify the [panda_moveit_config](https://github.com/ros-planning/panda_moveit_config) repository for this pull-request (contrary to what I do in my blog post), so the controller types for the gripper do not match:
The original package uses _GripperCommand_, Gazebo requires the _FollowJointTrajectory_ type controller.
As a consequence the gripper cannot be moved.


mehditlili:
Maybe it would be better to open a pull request for ros melodic branch directly.

erdalpekel:
Ok I'll base it on the melodic branch and see if changes are necessary there.

fwalch:
Thank you for the contribution, and sorry for the late response! We didn't take a proper look at this yet, but we'll try to do that soon. One thing I already can say: we want to make it clear that the given values are not the "real" values, but only to enable Gazebo. Maybe you can clarify this with something like `<!-- for simulation only, does not correspond to physical properties of robot -->` on the `inertia` blocks.

erdalpekel:
@fwalch That kind of comment is already included:
for simulation comment

mbreyer:
Are there any updates to incorporate this or any Gazebo support for the Panda arm in the future?

rickstaa:
@mbreyer I was wondering exactly the same thing? Maybe @fwalch knows more.

davetcoleman:
We're trying to add a Gazebo+MoveIt tutorial and would appreciate this being merged in:
https://github.com/ros-planning/moveit_tutorials/pull/499

Any updates?

erdalpekel:
Hi @davetcoleman,

I've just merged a pull request from [tahsinkose](https://github.com/tahsinkose) to my fork of franka_ros.
It works in ROS melodic.
I'm not sure what else to add on this PR.

Erdal

erdalpekel:
I attempted to fix the merge errors but it is not possible with the manual <collision> specifications that are currently in place.
Gazebo does not like the manual collision specifications between link0 and link1. The links push each other away in the simulation.
Also, the manual collision specifications on link8 and in the hand need to be removed because they cause problems with moveit.

When the stl mesh files are applied to these links for the collision model (this was the case when I initially forked the franka_ros repo last year) the simulation works again.

I think we need to distinguish between the simulation and real robot cases in the xacro model definition files and apply collision models accordingly.

peasant98:
Any updates on this @erdalpekel?

erdalpekel:
No, no updates from my side. I didn't get any feedback on my last comment.
My proposal was to distinguish simulation and real robot use in the config files.

tahsinkose:
I'd like to push forward this PR. First question: Do we really want to target `kinetic-devel` or should we port these changes to `melodic-devel`? Maybe both :thinking:

davetcoleman:
I just emailed @fwalch another request to merge this

tahsinkose:
@davetcoleman https://github.com/frankaemika/franka_ros/pull/126 is the actual PR. We need to merge that one.

fwalch:
Superseded by #126.

PR #79, opened 2019-10-14
Meshes for G8

marc-hanheide:
this is a small fix to have the Franka display correctly in gazebo8 (in Kinetic). Thanks to @pulver22 for finding this fix.


fwalch:
Thanks!

PR #110, opened 2020-08-03
Added New examples

Khesbak:
I developed two more interesting examples for pickup object and text writing using Cartesian controller examples

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,5 @@
 # ROS integration for Franka Emika research robots
 
-[![Build Status][travis-status]][travis]
-
 See the [Franka Control Interface (FCI) documentation][fci-docs] for more information.
 
 ## License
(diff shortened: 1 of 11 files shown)


gollth:
Hi @Khesbak,

thanks for the interest in our robot and `franka_ros` and sorry for the late reply! We welcome that you want to contribute here to make this software better. 

Looking at your controllers I have a few doubts on the usefulness of them, though. The `CartesianPickupTask` looks like the [CartesianPoseExampleController](https://github.com/frankaemika/franka_ros/blob/develop/franka_example_controllers/src/cartesian_pose_example_controller.cpp) wrapped inside a statemachine. The cartesian pose example controller already show cases how to use the cartesian pose interface of the robot to send position commands. Your example nicely shows how this approach can easily be integrated into a more sophisticated pick- & place-task. However I think usually you don't want to do the statemachine logic _inside_ your control loop (because of the realtime constraint) but rather employ a motion planning library (such as [Moveit](https://moveit.ros.org/)) or any other high level system _on top_ of the controller.

Also the parameters are hard coded into the C++ file, which also makes it not easily configurable (even worse, depending on the users environment, it might actually crash into something during execution =S)

Similar arguments hold for the `CartesianTextWritingTask`-Controller.

I'm happy for you that you were able to integrate these scenarios so easily, but very reluctant to include these in official examples, since I don't see what additional benefit they would offer.

PR #116, opened 2020-09-18
(franka-hw) add franka_control_services install instruction

jethrokuan:
Adds the missing install instruction for `franka_control_services`, which is now under the `franka-hw` package.

This is similar in spirit to #93, but combines the install instructions for `franka_hw` and `franka_control_services` like the other franka_ros packages.


sgabl:
Thanks for the fix!

PR #132, opened 2021-03-17
Removed panda_moveit_config dependency in franka_ros

c-rizz:
Hello, I have been installing franka_ros with ros noetic on ubuntu 20.04 and I noticed that the ros-noetic-franka-ros is not available on apt. By taking a quick look at jenkins (https://build.ros.org/job/Ndev__franka_ros__ubuntu_focal_amd64/4/console) it looks like the build is failing due to the panda_moveit_config dependency in franka_ros. This because panda_moveit_config is also not available in noetic. From what I understand this dependency is a remnant from when panda_moveit_config was a part of this repository, so it should be possible to safely remove it.

### DIFF
--- a/franka_ros/package.xml
+++ b/franka_ros/package.xml
@@ -15,7 +15,6 @@
 
   <exec_depend>franka_control</exec_depend>
   <exec_depend>franka_description</exec_depend>
-  <exec_depend>franka_example_controllers</exec_depend>
   <exec_depend>franka_gripper</exec_depend>
   <exec_depend>franka_hw</exec_depend>
   <exec_depend>franka_msgs</exec_depend>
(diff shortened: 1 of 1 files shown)


fwalch:
Hi,

it probably makes sense to remove the dependency, but that will not fix the build issues because `franka_example_controllers` also depends on `panda_moveit_config`. Related: https://github.com/ros-planning/panda_moveit_config/issues/77

Probably we could remove `franka_example_controllers` from `franka_ros` as well, as its only an example package.. can you add this to your PR?

Thanks!

c-rizz:
Done, just removed the dependency in package.xml

fwalch:
Thanks!

PR #144, opened 2021-08-11
Fixes missing example controllers dependency

rickstaa:
As, some scripts use the uses the moveit_commander (e.g. [move_to_start.py](https://github.com/frankaemika/franka_ros/blob/356ec114c488c75edeaf5315633359459d37af31/franka_example_controllers/scripts/move_to_start.py#L4)) it should be added as a ROS dependency.


gollth (changes requested):
Makes sense. Can you again rebase to our latest develop and solve the conflict please?

rickstaa:
@gollth Done

gollth (approved):

PR #146, opened 2021-08-13
Adds accurate finger collision mesh

rickstaa:
This commit replaces the more bold finger collision mesh with the more accurate version that is found in the
[AndrejOrsula/panda_ign](https://github.com/AndrejOrsula/panda_ign/blob/master/panda/meshes/collision/finger.stl) repository.


rickstaa:
This pull request is also included in #147. As that change requires the commit in this pull request. I will therefore close this pull request since I think it is better to merge #147. Feel free to reopen and merge if you need this pull request.

PR #151, opened 2021-08-18
Moves finger collision geometries to the finger joints

rickstaa:
This commit moves the finger collision geometries to the finger joints. This was done to prevent errors in third party libraries. Moveit for example expects a collision geometry in each joint that can be controlled. You can see this behaviour by running the `panda_moveit_config demo.launch` launch file. If the old description is used moveIt throws the following errors:

```bash
[ WARN] [1629295377.872753565]: Link panda_leftfinger has visual geometry but no collision geometry. Collision geometry will be left empty. Fix your URDF file by explicitly specifying collision geometry.
[ WARN] [1629295377.872794348]: Link panda_rightfinger has visual geometry but no collision geometry. Collision geometry will be left empty. Fix your URDF file by explicitly specifying collision geometry.
```

##### Safety distance 0.03

*Open state*:

image

*Closed state*:

image

##### Safety distance 0.0

image

##### Safety distance 0.03

image

##### Safety distance 0.0

image

### DIFF
--- a/franka_description/robots/hand.xacro
+++ b/franka_description/robots/hand.xacro
@@ -33,24 +33,6 @@
           <sphere radius="${0.04+safety_distance}"  />
         </geometry>
       </collision>
-      <collision>
-        <origin xyz="0 0 0.1" rpy="0 ${pi/2} ${pi/2}"/>
-        <geometry>
-          <cylinder radius="${0.02+safety_distance}" length="0.1" />
-        </geometry>
-      </collision>
-      <collision>
-        <origin xyz="0 -0.05 0.1" rpy="0 0 0"/>
-        <geometry>
-          <sphere radius="${0.02+safety_distance}"  />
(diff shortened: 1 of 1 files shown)


rhaschke:
I agree with @rickstaa that the collision volumes of the fingers should be attached to the finger links, but not to the hand link.
As far as I understand the coarse collision volumes (from spheres and cylinders) are the ones used internally for self-collision checking. As such, they shouldn't affect too much collision checking with the environment. However, particularly the hand collision models are extremely coarse:
image

I suggest using thin cylinders to model the fingers. Also, the big cylinder around the hand base should be decreased in size, such that the collision geometry can be employed to actually plan grasps! Currently, one would need to disable collision for the full hand to allow grasping of objects, which doesn't increase safety!

rickstaa:
Closed as this is implemented in #188.

PR #152, opened 2021-08-18
move finger collision geometry sparse

rickstaa:
Similar to #151 but now the cylinder geometry between the two fingers have been removed.

##### Safety distance 0.03

image

##### Safety distance 0.0

image

##### Safety distance 0.03

image

##### Safety distance 0.0

image


rickstaa:
Closed since this is implemented in #189

PR #153, opened 2021-08-18
Adds the ability to specify the safety_distance

rickstaa:
This pull request gives users the ability to specify the `safety_distance` when they load the robot xacro. This only changes the safety_distance in the `urdf` and does not affect the `safety_distance` in the internal controller.


rickstaa:
As discussed in our meeting this will be handled in another pull request. Closing this for now.

PR #154, opened 2021-08-19
Adds 'use_collision_mesh' arg to the robot description

rickstaa:
This commits gives users the ability to load the `robot_description` `urdf` with the collision meshes instead of the collision geometries. This is for example useful if people want to use less strict Mesh-based collision properties for third party planners in their packages.

For the [moveit_tutorials](https://github.com/ros-planning/moveit_tutorials) for example, the movement with the geometry messes is to course (see https://github.com/ros-planning/panda_moveit_config/issues/72#issuecomment-768472329). In that case, or the geometries need to be improved or we might want to use the meshes instead.

### DIFF
--- a/franka_description/robots/hand.xacro
+++ b/franka_description/robots/hand.xacro
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <robot xmlns:xacro="http://www.ros.org/wiki/xacro" name="hand">
   <!-- safety_distance: Minimum safety distance in [m] by which the collision volumes are expanded and which is enforced during robot motions -->
-  <xacro:macro name="hand" params="connected_to:='' ns:='' rpy:='0 0 0' xyz:='0 0 0' safety_distance:=0">
+  <xacro:macro name="hand" params="connected_to:='' ns:='' rpy:='0 0 0' xyz:='0 0 0' safety_distance:=0 use_collision_mesh:=false">
     <xacro:unless value="${connected_to == ''}">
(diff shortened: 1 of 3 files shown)


rickstaa:
I can of course also change the name of the variable to, for example, `strict_collisions`.

rhaschke (commented):
:+1: As a user, I definitely want to have an option to resort to these fine-grained meshes.
Is there an option to disable the internal self-collision checking of the robot controller (and only rely on planning methods)?

rickstaa:
@rhaschke Do we want to keep this open? It is related to #189 but not fully similar since for the arm still geometries are used. I think #189 is good enough for the `panda_moveit_config` and my own experiments.

rhaschke:
As Franka pointed out that the internal self-collision checking cannot be disabled or changed by any means, it doesn't make sense to plan with the detailed meshes. However, for _environmental_ collisions, the same problem arises as for the hand: While planning might fail with the coarse self-collision models, the robot might perfectly fit into a narrow passage with the detailed meshes. 
Thus, actually, #189 should be extended to apply the same idea to the whole arm, i.e. define detailed meshes for environmental collisions and use separate coarse models for the self-collision. Instead of the name `coarse`, we should use another suffix, e.g. `sc` to indicate their use for self-collision checking.

I think, this PR here, can be closed.

rickstaa:
Makes sense. Do you want to extend #189 or create a separate pull request?

rhaschke:
See #199 :wink:

PR #155, opened 2021-08-23
Fixes hand real/gazebo discrepacy

rickstaa:
This commit makes sure that the hand joint control behaviour between the real and simulated robot is equal. Previously the 'franka_finger_joint2' mimics the 'franka_finger_joint1' in the real robot but not in the simulation.

I think having the same behaviour in both the simulation and on the real robot makes sense since it otherwise might causes errors in third party packages controlling the robot.


rickstaa:
This appears to break the [panda_moveit_config](https://github.com/ros-planning/panda_moveit_config) package. I will investigate if this change can be applied while using MoveIt.

rickstaa:
Closed, since adding the `mimic` tag to the gazebo `urdf` is currently not supported (see https://github.com/ros-simulation/gazebo_ros_pkgs/issues/251).

PR #171, opened 2021-09-27
Fixes franka_gripper_sim GripperCommand action

rickstaa:
This commit makes sure that users can also send gripper widths between `0.0` and `0.08`. In the old version the gripper would only open and close based on whether the set gripper width was smaller or bigger than the current width.

I explained the problem in #172.


rickstaa:
Closed because I want this to be on a different branch.

Issue #172, opened 2021-09-27
[bug] Gazebo gripper control is broken


rickstaa:
<details>
<summary>Old bug report</summary>

In the current version, when working with the gazebo simulation, the gripper width that is sent to the `/panda/franka_gripper/gripper_action/goal` topic is not handled correctly. The problem is caused by the following code:

https://github.com/frankaemika/franka_ros/blob/054af531f43f3fc1ddb439cc4102a41f87a8e0da/franka_gazebo/src/franka_gripper_sim.cpp#L436

Due to this, the desired gripper with now becomes `0` when the gripper width is smaller than the current gripper width and `0.08` when it is bigger than the current gripper width. Consequently, users can now only open/close the gripper. 

I created #173 to address this problem.

#### Steps to reproduce

1. Follow [the documentation](https://frankaemika.github.io/docs/installation_linux.html) to install `libfranka` library.
2. Create a new catkin workspace.
3. Clone [franka_ros](https://github.com/frankaemika/franka_ros) package in this workspace.
4. Clone https://github.com/rickstaa/panda_moveit_config/tree/adds_gazebo_simulation in this workspace.
5. Build the workspace following the steps explained in [the documentation](https://frankaemika.github.io/docs/installation_linux.html#building-the-ros-packages).
6. Start the Panda simulation using the `roslaunch panda_moveit_config demo_gazebo.launch`.
8. Set the ros namespace `export ROS_NAMESPACE=/panda
7. Run the script below to send a MoveIt gripper command.
8. See that you can not set the gripper to a specific width, since it will only close and open.

```python
"""Small script to see that the gazebo GripperCommand action does not set the gripper
width.
"""

import sys

import moveit_commander
import moveit_msgs.msg
import rospy

GRIPPER_WIDTH = 0.08

if __name__ == "__main__":

    # Initiate MoveIt commander and node
    moveit_commander.roscpp_initialize(sys.argv)
    rospy.init_node("test_gripper_width", anonymous=True)

    # Create commanders
    robot = moveit_commander.RobotCommander()
    scene = moveit_commander.PlanningSceneInterface()

    # Load arm and hand groups
    hand_move_group = moveit_commander.MoveGroupCommander("hand")

    # Display the trajectory
    display_trajectory_publisher = rospy.Publisher(
        "/move_group/display_planned_path",
        moveit_msgs.msg.DisplayTrajectory,
        queue_size=20,
    )

    # Send the gripper width command
[...]

Issue #161, opened 2021-08-30
[feature_request] Adds ability to switch gazebo hardware interface


rickstaa:
I think it would be handy to have the ability to switch the `hardware_interface` that is used in the gazebo simulation for controlling the panda arm. This feature would allow users to test control the robot using different hardware interfaces, similar to the real robot. 

Currently,  the `hardware_interface/JointEffortInterface` is hardcoded into the [panda_arm.urdf.xacro](https://github.com/frankaemika/franka_ros/blob/356ec114c488c75edeaf5315633359459d37af31/franka_description/robots/panda_arm.urdf.xacro#L34-L35) file. Because of this, the release of the Noetic [panda_moveit_config](https://github.com/ros-planning/panda_moveit_config) package, which targets franka_ros v0.8.0, is blocked, as we would like to both use the `JointPositionInterface` in the gazebo simulation and on the real robot (see https://github.com/ros-planning/panda_moveit_config/pull/74 and https://github.com/ros-planning/panda_moveit_config/pull/68).

I also tried switching to the `effort_controllers/JointTrajectoryController` while using the following pid gains:

```yaml
effort_joint_trajectory_controller:
  type: effort_controllers/JointTrajectoryController
  joints:
    - panda_joint1
    - panda_joint2
    - panda_joint3
    - panda_joint4
    - panda_joint5
    - panda_joint6
    - panda_joint7
  gains:
    panda_joint1: { p: 12000, d: 50, i: 0.0, i_clamp: 10000 }
    panda_joint2: { p: 30000, d: 100, i: 0.02, i_clamp: 10000 }
    panda_joint3: { p: 18000, d: 50, i: 0.01, i_clamp: 1 }
    panda_joint4: { p: 18000, d: 70, i: 0.01, i_clamp: 10000 }
    panda_joint5: { p: 12000, d: 70, i: 0.01, i_clamp: 1 }
    panda_joint6: { p: 7000, d: 50, i: 0.01, i_clamp: 1 }
    panda_joint7: { p: 2000, d: 20, i: 0.0, i_clamp: 1 }
```

Unfortunately, although these gains seem decent on the simulation robot (i.e. control has a slight offset), they lead to dangerous situations (i.e. very high velocities and acceleration) on the real robot, which causes the software emergency stop to be invoked).

I created #156 and #157 to add this feature.

rickstaa:
[...]

PR #173, opened 2021-09-27
Fixes franka_gripper_sim GripperCommand action

rickstaa:
This commit makes sure that users can also send gripper widths between `0.0` and `0.08`. In the old version the gripper would only open and close based on whether the set gripper width was smaller or bigger than the current width.

I explained the problem in #172.

### DIFF
--- a/franka_gazebo/src/franka_gripper_sim.cpp
+++ b/franka_gazebo/src/franka_gripper_sim.cpp
@@ -421,22 +421,35 @@ void FrankaGripperSim::onGraspGoal(const franka_gripper::GraspGoalConstPtr& goal
 }
 
 void FrankaGripperSim::onGripperActionGoal(const control_msgs::GripperCommandGoalConstPtr& goal) {
+  control_msgs::GripperCommandResult result;
+  double width_d = goal->command.position * 2.0;
+
   ROS_INFO_STREAM_NAMED("FrankaGripperSim", "New Gripper Command Action Goal received: "
-                                                << goal->command.max_effort << "N");
+                                                << width_d << "m, " << goal->command.max_effort
+                                                << "N");
 
   // HACK: As one gripper finger is <mimic>, MoveIt!'s trajectory execution manager
   // only sends us the width of one finger. Multiply by 2 to get the intended width.
   double width = this->finger1_.getPosition() + this->finger2_.getPosition();
 
+  if (width_d > kMaxFingerWidth || width_d < 0.0) {
+    std::string error =
+        "Commanding out of range width! max_width = " + std::to_string(kMaxFingerWidth) +
+        " command = " + std::to_string(width_d);
+    ROS_ERROR_STREAM_NAMED("FrankaGripperSim", error);
+    result.reached_goal = static_cast<decltype(result.reached_goal)>(false);
+    action_gc_->setAborted(result, error);
+    return;
+  }
+
   franka_gripper::GraspEpsilon eps;
   eps.inner = this->tolerance_gripper_action_;
   eps.outer = this->tolerance_gripper_action_;
 
-  transition(State::GRASPING,
-             Config{.width_desired = goal->command.position * 2.0 < width ? 0 : kMaxFingerWidth,
-                    .speed_desired = this->speed_default_,
-                    .force_desired = goal->command.max_effort,
-                    .tolerance = eps});
+  transition(State::GRASPING, Config{.width_desired = width_d,
+                                     .speed_desired = this->speed_default_,
+                                     .force_desired = goal->command.max_effort,
+                                     .tolerance = eps});
 
   waitUntilStateChange();
 
@@ -446,7 +459,6 @@ void FrankaGripperSim::onGripperActionGoal(const control_msgs::GripperCommandGoa
     return;
   }
 
-  control_msgs::GripperCommandResult result;
   if (this->state_ != State::HOLDING) {
     result.reached_goal = static_cast<decltype(result.reached_goal)>(false);
     std::string error = "Unexpected state transistion: The gripper not in HOLDING as expected";
@@ -454,7 +466,6 @@ void FrankaGripperSim::onGripperActionGoal(const control_msgs::GripperCommandGoa
     return;
   }
 
-  double width_d = goal->command.position * 2.0;
   width = this->finger1_.getPosition() + this->finger2_.getPosition();  // recalculate
   bool inside_tolerance = width_d - this->tolerance_gripper_action_ < width and
                           width < width_d + this->tolerance_gripper_action_;
--- a/franka_gazebo/config/franka_hw_sim.yaml
+++ b/franka_gazebo/config/franka_hw_sim.yaml
@@ -1,9 +1,9 @@
 arm_id: $(arg arm_id)
 
-m_ee:  0.76
+m_ee: 0.76
 franka_gripper:
-  type:    franka_gazebo/FrankaGripperSim
-  arm_id:  $(arg arm_id)
+  type: franka_gazebo/FrankaGripperSim
+  arm_id: $(arg arm_id)
 
   finger1:
     gains: { p: 100, i: 25, d: 20 }


rickstaa:
Please check the `MoveIt gripper control execution timeout` section of this comment https://github.com/frankaemika/franka_ros/issues/161#issuecomment-927969666 before merging this commit. Although, this comment does not directly affect this pull request, I think it is good to be aware of issues that exist with the current implementation of the `GRASP` state.

franka_gazebo/src/franka_gripper_sim.cpp
```diff
@@ -422,21 +422,33 @@ void FrankaGripperSim::onGraspGoal(const franka_gripper::GraspGoalConstPtr& goal
...
+
   // HACK: As one gripper finger is <mimic>, MoveIt!'s trajectory execution manager
   // only sends us the width of one finger. Multiply by 2 to get the intended width.
   double width = this->finger1_.getPosition() + this->finger2_.getPosition();
+  double width_d = goal->command.position * 2.0;
+
+  if (width_d > kMaxFingerWidth || width_d < 0.0) {
+    ROS_ERROR_STREAM_NAMED("FrankaGripperSim", "Commanding out of range width! max_width = "
+                                                   << kMaxFingerWidth << " command = " << width_d);
+    std::string error = "Commanding out of range width";
```
gollth: Can you please use the same error message here for the `ROS_ERROR` and abortion message please?

franka_gazebo/src/franka_gripper_sim.cpp
```diff
@@ -422,21 +422,33 @@ void FrankaGripperSim::onGraspGoal(const franka_gripper::GraspGoalConstPtr& goal
...
   ROS_INFO_STREAM_NAMED("FrankaGripperSim", "New Gripper Command Action Goal received: "
+                                                << goal->command.position * 2.0 << "m, "
                                                 << goal->command.max_effort << "N");
 
+  control_msgs::GripperCommandResult result;
+
   // HACK: As one gripper finger is <mimic>, MoveIt!'s trajectory execution manager
   // only sends us the width of one finger. Multiply by 2 to get the intended width.
   double width = this->finger1_.getPosition() + this->finger2_.getPosition();
+  double width_d = goal->command.position * 2.0;
```
gollth: If you move this line to the top of the function, you can safe the recalculation in line 425

franka_gazebo/src/franka_gripper_sim.cpp
```diff
@@ -422,21 +422,33 @@ void FrankaGripperSim::onGraspGoal(const franka_gripper::GraspGoalConstPtr& goal
...
   franka_gripper::GraspEpsilon eps;
   eps.inner = this->tolerance_gripper_action_;
   eps.outer = this->tolerance_gripper_action_;
 
-  transition(State::GRASPING,
-             Config{.width_desired = goal->command.position * 2.0 < width ? 0 : kMaxFingerWidth,
-                    .speed_desired = this->speed_default_,
-                    .force_desired = goal->command.max_effort,
-                    .tolerance = eps});
+  transition(State::GRASPING, Config{.width_desired = goal->command.position * 2.0,
```
gollth: Here you could use `width_d` instead of recalculating

gollth (changes requested):

rickstaa:
@gollth I applied the changes you requested. I'm now looking at the comments you made about the PID gains in #172.

rickstaa:
@gollth I also found PID gains similar to yours. I applied yours on this pull request. I think the `hardware_interface::EffortJointInterface` behaves close enough to a 1st order system that a derivative gain is not needed to prevent overshoot.

rickstaa:
@gollth I went ahead and apply your suggestion in #172 for the change of namespace. If you want, we can merge https://github.com/rickstaa/franka_ros/pull/6 into #173 or create a new pull request. The related doc change can be found in https://github.com/frankaemika/docs/pull/13.

rickstaa:
@gollth There still appears to be an issue with the state machine when the PID gains you gave in #172. It still does not transfer from the `GRASP` to the `HOLDING` state because the velocities are sometimes higher than the current `kGraspRestingThreshold `(i.e. `0.001`). Setting the `kGraspRestingThreshold` to `0.005` seems to solve the problem.

https://github.com/frankaemika/franka_ros/blob/0a9563c0addfe0bc212013a7702f40566163238c/franka_gazebo/include/franka_gazebo/franka_gripper_sim.h#L36

velocities

In the related comment (https://github.com/frankaemika/franka_ros/issues/172#issuecomment-932230733), during testing you use the  [config/franka_hw_sim.yaml](https://github.com/frankaemika/franka_ros/blob/0a9563c0addfe0bc212013a7702f40566163238c/franka_gazebo/config/franka_hw_sim.yaml) to increase both the `resting_threshold` and `consecutive_samples` values. Due to that I was not sure whether you wanted to update the [franka_gripper_sim](https://github.com/frankaemika/franka_ros/blob/0a9563c0addfe0bc212013a7702f40566163238c/franka_gazebo/include/franka_gazebo/franka_gripper_sim.h) `kGraspRestingThreshold` parameter or the [config/franka_hw_sim.yaml](https://github.com/frankaemika/franka_ros/blob/0a9563c0addfe0bc212013a7702f40566163238c/franka_gazebo/config/franka_hw_sim.yaml).

Please let me know where you want to add these values or apply them yourself using `gh pr checkout 173`.

rickstaa:
I did one last investigation into the behaviour under different PID gains. I did this both for the `move` and `gripper_action` action services. The results can be found in the next section.

## Gripper_action step response

#### Old PID gains

OLD_PID_GRIPPER_ACTION_STEP

#### New PID gains

NEW_PID_GRIPPER_ACTION_STEP

## Move action sinusoid response

#### Old PID gains

OLD_PID_MOVE_ACTION_SIN

#### New PID gains

NEW_PID_MOVE_SIN

## Conclusion

From these experiments, you can see that although with the new PID gains the resting velocities are lower, the overall behaviour is a lot worse. This is expected since no derivative gain is added to smooth out the motion and no integral gain to get rid of a steady-state error. Unfortunately, I therefore think the old gains are closer to the optimal gains.

For the state machine to transfer from `GRASPING` to `HOLDING` with the old PID gains, you however need to set a `kGraspRestingThreshold` of `0.11` which is not close to being stationary.

## Further tuning

I tried to tune the gains manually again, but I keep getting gains close to the old ones (i.e., `P: 100-500, I: 30-60, D: 10-22`). The old gains (i.e., `100-25-20` seem to give good smooth control results, but sadly high velocity oscillations. I however can not get rid of the oscillations without making the movement less smooth. 

Furthermore, I also tried using MATLAB to tune the PID gains (see https://www.youtube.com/watch?v=ENhFepqjFOY and the zip below) this however gives me a wide range of gains. I think this is since I'm not directly controlling the joint, but sending commands to the action server. As a result a lot of the control commands get cancelled.

[gripper_pid_tunner.zip](https://github.com/frankaemika/franka_ros/files/7294637/gripper_pid_tunner.zip)

rickstaa:
@gollth To ensure the velocities are not resulting from the arm controller, I also checked the old PID gains while running the `cartesian_impedance` controller (see GIF below). At this stage, I still think the problem we are experiencing has one of the following causes (ordered from most likely to least likely):

- Incorrect PID gains.
- Incorrect dynamic finger properties (i.e. inertias, masses, joint friction).
- Upstream issues in the gazebo simulator.

impedance_controller

### Incorrect PID gains

For my investigation, see https://github.com/frankaemika/franka_ros/pull/173#issuecomment-933435669.

### Incorrect dynamics

I quickly investigated the effect of the dynamic parameters.

| property | Effect                                                                           |
|----------|----------------------------------------------------------------------------------|
| Damping  | Increasing damping increases vel oscillations. Lowering has no noticeable effect. |
| Friction | Increasing friction lowers vel oscillations but prevents movement.               |
| Inertia  | Doesn't affect the vel oscillations too much (i.e., for the values I tested).     |
| Mass     | The vel oscillations are lowered when the mass is increased.                     |

I interchanged the inertias you provided by the ones used in the [justagist/franka_panda_description](https://github.com/justagist/franka_panda_description/blob/62335618e7f141755524836c738efe9c01396440/robots/hand.xacro#L58-L62) repository. If we keep the mass the same (i.e. `15g`), we see no significant effect on the velocity oscillations. If we, however, increase the mass to `100` g, the velocity oscillations decrease.

To me, `100` g looks a bit much, where `15` g seems too little. However, I do not know the exact weight since I have not detached the finger from the hand and can not find the hand data sheet online (see the results for `38` g in the gif below). 

To conclude, I think the oscillations are caused by an interplay between the PID gains and the dynamic properties. Maybe you can  verify the finger weight such that we can try to retune the PID gains again. If that doesn't help, we can create a question on the [ROS](https://answers.ros.org/questions/) or [gazebo](https://answers.gazebosim.org/questions/) forms to see if somebody encountered this behaviour on other robots.

38 g mass

### Bugs in gazebo

 I checked the gazebo repo but only found issues with the velocity calculation when a joint is in contact with an object (i.e. https://github.com/osrf/gazebo/issues/622 and https://github.com/osrf/gazebo/issues/634). Further, given the fact that we can actually see the fingers shake, I don't think this problem is caused by a bug in Gazebo.

finger_shaking

marcbone:
The current implementation of the grasp command in libfranka also opens or closes the gripper completely. Therefore, we do not want merge this, as it does not reflect the real behavior of the gripper.

rickstaa:
@marcbone Thanks for your comment. I see what you mean the `gripper.grasp` method gets called after the `gripper.move` therefore invoking the grasp behaviour implemented in the libfranka library (see https://github.com/frankaemika/libfranka/blob/f1f46fb008a37eb0d1dba00c971ff7e5a7bfbfd3/src/gripper.cpp#L73-L74):

https://github.com/frankaemika/franka_ros/blob/ac1e07d4d7e2a916dbf8b3a10a7fc73277a08fc2/franka_gripper/src/franka_gripper.cpp#L54-L55

In that case, closing this pull request makes sense since we want the simulation to mimic the actual robot behaviour. However, there is one downside with this: users can not set a specific with using MoveIt. They have to use the `franka_gripper/move` action instead. 

Ideally, there would be another `franka_gripper/move` action (e.g. `franka_gripper/moveit_move`) that has the `control_msgs/GripperCommand` msg-type. MoveIt could then use this action server to set the gripper width to a specific width. 

@rhaschke Do you think not having the ability to set a specific width using MoveIt is a significant problem?

rhaschke:
@marcbone, so the real hand doesn't allow to drive the gripper to a specific position (different from fully opened / closed)?
If that's not possible, I would consider this as a severe limitation of the gripper and asking for a this feature!

In any case, the simulation should, of course, reflect real robot behavior.

rickstaa:
@rhaschke On the real robot, I normally set the gripper to a specific position using the [franka_gripper::MoveAction(width, speed)](https://frankaemika.github.io/docs/franka_ros.html#franka-gripper) action. This action is found on the `franka_gripper/move` namespace (see [the docs](https://frankaemika.github.io/docs/franka_ros.html#franka-gripper0). 

The main problem is that MoveIt, to my knowledge, only works with the [control_msgs.msg.GripperCommand](https://docs.ros.org/en/api/control_msgs/html/action/GripperCommand.html) msg while the `franka_gripper/move` service uses the `franka_gripper/Move` message.

https://github.com/frankaemika/franka_ros/blob/ac1e07d4d7e2a916dbf8b3a10a7fc73277a08fc2/franka_gripper/action/Move.action#L1-L6

As a result, people can not set the gripper width using MoveIt since the `franka_gripper/gripper_action` transitions into the `grasp` state and therefore only opens/closes the gripper. They have to use the `franka_gripper/move` action service directly to do this.

rhaschke:
Indeed, MoveIt can only interact with `GripperCommand` actions or `TrajectoryJointControllers`. If Franka provides another action type, a proxy would be needed to translate corresponding actions.

rhaschke:
Is there a trajectory controller available for the gripper?

marcbone:
We can offer you to change the gripper command so that if the max_effort is 0 we use the gripper.move() command instead. Would that work for you? You could then move gripper to a desired width and would still be able to grasp an object with a certain force.

There is no trajectory controller available for the gripper since all gripper commands are blocking.

rhaschke:
That sounds reasonable.

rickstaa:
@marcbone I think your proposed solution is more intuitive and gives us a way to implement both the `move` and `grasp` behaviour at the MoveIt side. 

@rhaschke When @marcbone implements his solution. Until we implement a solution at our end I think we have to increase the `max_effort` that is set in the [moveit_simple_controller_manager](https://github.com/ros-planning/moveit/tree/master/moveit_plugins/moveit_simple_controller_manager) to a higher number to keep the current grasping behaviour (see https://github.com/ros-planning/moveit/blob/4db626d9c3b5d6296b012188a4a0bfe4bddf6bce/moveit_plugins/moveit_simple_controller_manager/include/moveit_simple_controller_manager/gripper_controller_handle.h#L111-L113).

rickstaa:
I will open an issue on https://github.com/ros-planning/moveit when @marcbone solution has been merged into the `devel` branch.  @marcbone can you maybe close https://github.com/frankaemika/franka_ros/issues/172 when you implemented your proposed solution. This way I know when we can start development at the MoveIt side.

rhaschke:
@rickstaa, I think you can start development at the MoveIt side already. Currently, a non-zero gripper effort can be specified [via the trajectory](https://github.com/ros-planning/moveit/blob/4db626d9c3b5d6296b012188a4a0bfe4bddf6bce/moveit_plugins/moveit_simple_controller_manager/include/moveit_simple_controller_manager/gripper_controller_handle.h#L132-L133). I think we should add the possibility to
- specify that as a ROS parameter in the GripperCommand declaration
- allow switching controllers for different (types of) trajectories
  https://github.com/ros-planning/moveit/pull/1832

rickstaa:
@rhaschke I created https://github.com/ros-planning/moveit/issues/2956. I can not give an ETA when I have time to implement this but please add me under assignees so that I don't forget.

rhaschke:
> We can offer to change the gripper command so that if the max_effort is 0 we use the gripper.move() command instead.

Thinking about the issue a little deeper, I think that your GripperAction misbehaves. Please correct me, @marcbone, if I'm wrong.
The `GripperCommand` msg provides two fields:
```
float64 position
float64 max_effort
```
As far as I understand, the current implementation kind of ignores the position argument of that message, but either attempts to fully open or fully close the hand depending on whether the current position is smaller or larger than the commanded position.
That's a reasonable approach for grasping. However, it doesn't apply to free-space motion, but exactly leads to the observed oscillations around the commanded position (https://github.com/frankaemika/franka_ros/pull/173#issuecomment-938507534).
IMO, the gripper action should realize _position-control_ as long as measured efforts are smaller than `max_effort`. Otherwise, it should switch to _force-control_ and maintain the given (grasp) force.

gollth:
> There still appears to be an issue with the state machine when the PID gains you gave in #172.

For the record, it was really a problem with poorly tuned gains for the gripper sim. With drastically reduced I + D gains the behavior is much more stable:

gripper-grasp-opening_

gollth:
> IMO, the gripper action should realize _position-control_ as long as measured efforts are smaller than `max_effort`. Otherwise, it should switch to _force-control_ and maintain the given (grasp) force.

Good idea @rhaschke! However we tried to stick as close to the implementation of the `franka_gripper` as possible.

rickstaa:
@gollth I will answer https://github.com/frankaemika/franka_ros/issues/172#issuecomment-976841002 here to keep the discussion in one place.

> > In the current form, the gripper action still cannot grasp objects (see gif).
> 
> This is kind of expected when you pass a gripper command position of 0.0. When you check the result of the action or the [documentation of franka_gripper](https://frankaemika.github.io/docs/franka_ros.html#franka-gripper) you will find out why. For the same reason, grasping works in your example above.

I indeed expected the `gripper_action` to fail based on the current codebase. As you are aware, it fails is caused by the ternary expression, which I tried to address in #173:

https://github.com/frankaemika/franka_ros/blob/414bcb1d3acf013e472586fce1404365f352eb9d/franka_gazebo/src/franka_gripper_sim.cpp#L437

As discussed earlier, this ternary expression makes it so that MoveIt can only `open` or `close` the gripper through the `gripper_action` service. As a result, the width is set to `0.0` when closing the gripper, and the following code is executed because the gripper width is not within the epsilon range:

https://github.com/frankaemika/franka_ros/blob/414bcb1d3acf013e472586fce1404365f352eb9d/franka_gazebo/src/franka_gripper_sim.cpp#L467-L468

This behaviour can also be observed in the `grasp` service when we set the width to `0.0` while keeping the epsilon at `0.005`.

```bash
rostopic pub --once /franka_gripper/grasp/goal              franka_gripper/GraspActionGoal              "goal: { width: 0.0, epsilon:{ inner: 0.005, outer: 0.005 }, speed: 0.1, force: 5.0}"
```  

> Good idea @rhaschke! However, we tried to stick as close to the implementation of the franka_gripper as possible.

I fully understand that you want to stick as close to implementing the real `franka_gripper` as possible, and I also think that the `grasp` action is implemented correctly. I, however, think the `gripper_action` in the current form is not very useful on the real and simulated robot when looking from the MoveIt perspective. This is because it does not allow us to set a specific with or grasp an object.

I think @marcbone's [suggestion](https://github.com/frankaemika/franka_ros/pull/173#issuecomment-964994857) together with https://github.com/ros-planning/moveit/issues/2956 would solve the `gripper_width` problem for which I created #172.

> We can offer you to change the gripper command so that if the max_effort is 0 we use the gripper.move() command instead. Would that work for you? You could then move the gripper to the desired width and would still be able to grasp an object with a certain force.
> 
> There is no trajectory controller available for the gripper since all gripper commands are blocking.

Additionally, @rhaschke suggestion would allow us also to grasp objects. I will leave the actual implementation to you, but I just wanted to sketch which problems we are experiencing from the MoveIt side.

rickstaa:
I just noticed that we could make MoveIt grasp by setting the `/franka_gripper/gripper_action/width_tolerance` to the gripper width (i.e. `0.08`). I, however, still think that the gripper_action is a bit unintuitive and could be improved. The downside is that using this ROS parameter, we can OR grasp OR move since it is only read at startup:

https://github.com/frankaemika/franka_ros/blob/414bcb1d3acf013e472586fce1404365f352eb9d/franka_gazebo/src/franka_gripper_sim.cpp#L29-L30

Together with @marcbone suggestion, this would allow MoveIt to both move and grasp. @rhaschke are you okay with using that ROS parameter for MoveIt?

rickstaa:
Correction, it appears that setting the `/franka_gripper/gripper_action/width_tolerance` does not work when it is set to `0.08` since the check is performed too early when the gripper width is still in the start position:

https://github.com/frankaemika/franka_ros/blob/414bcb1d3acf013e472586fce1404365f352eb9d/franka_gazebo/src/franka_gripper_sim.cpp#L466-L469

 Setting it to a value greater than the gripper width (i.e. `0.1`) works but feels like a rather peculiar workaround. Further, since the direction is not included in the GripperAction `STATE::GRASPING` transition:

https://github.com/frankaemika/franka_ros/blob/414bcb1d3acf013e472586fce1404365f352eb9d/franka_gazebo/src/franka_gripper_sim.cpp#L439

We have to set the max_effort to be negative:

```bash
rostopic pub --once /franka_gripper/gripper_action/goal control_msgs/GripperCommandActionGoal "goal: {command:{position: 0.0, max_effort: -5.0}}"
```

rickstaa:
@gollth, @marcbone, @rhaschke maybe we can continue our discussion in #172?

gollth:
Why adjust the tolerance instead of the actual desired grasping position? When you want to plan a grasping motion you should have an idea of the size you want to grasp no?

rickstaa:
> Why adjust the tolerance instead of the actual desired grasping position?

@gollth Oh, I see. Sorry for being unclear in my previous comment. My comment above was about how I can hack the `gripper_action` so that it supports both movement and grasping. That is why I adjusted the `width_tolerance` since the ternary expression prohibits me from setting a specific width.

> When you want to plan a grasping motion you should have an idea of the size you want to grasp no?

This is true when you have information about the grasping object, and you can use this information to get a grasping Pose. I also have use-cases where I don't care about the object width, but I want the gripper to close while making sure that it keeps applying a force without the force exceeding the `max_effort`. A mechanism, as explained by @rhaschke earlier, would help in this case.

To conclude, I think the current implementation of the grasping service is perfect since it allows the user to specify a specific grasp. I, however, think that the `grasp_action` service (on both the real and simulated robot), which MoveIt uses should allow the following behaviours:

- When you send a `gripper_width`, and a `max_effort` of `0.0`, the move_action should be used.
- When you send a `gripper_width` and a `max_effort` that is greater than `0.0`, the gripper should try to go to the `gripper_width` and only move into the `HOLDING` state when the effort exceeds the `max_effort`.

By doing this, users can now both move and grasp in MoveIt (on the real and simulated robot). I also think this makes more sense given the structure of the [GripperCommand](https://docs.ros.org/en/api/control_msgs/html/msg/GripperCommand.html) message. I also don't think this change removes functionalities since you already supply a `grasping` and `move` service.

rickstaa:
> > There still appears to be an issue with the state machine when the PID gains you gave in #172.
> 
> For the record, it was a problem with poorly tuned gains for the gripper sim. With drastically reduced I + D gains, the behaviour is much more stable:
> 
> gripper-grasp-opening_

@gollth, I tested your new gains, and I'm sorry, but they do not work correctly on the simulated robot. They seem to work when the hand is horizontal since there is no gravity being applied to the fingers. However, when I turn the gripper by some degrees (i.e. 90 in the example below), the low gains cannot move the fingers up (see gif 1).

**gif 1:**
wrong_gains

I improved the behaviour by increasing both the P, I and D values again (see gif 2/3). The velocities now stay in the `0.004` range when the gripper is vertical and `0.0002` when the gripper is horizontal. I did not spend more than 5 min tuning these gains, so there are probably better gains that can decrease the velocity variance while still giving the correct behaviour. I, however, hope my gains serve as a starting point to find the solution to this very delicate tuning problem.

**gif 2:**
better_gains

**gif 3:**
velocities

rickstaa:
@gollth I just realized that I checked the `move` service while the problems are only experienced when using the `grasp` service. The issues are still present when using the grasp service (see gif below).

grasp_velocity

It is, however strange that these velocities are not present when using the moving service. The oscillations are, therefore, closely related to the logic you use for the GRASP state. This is because the current logic assumes that the gripper comes into contact with another object which helps dampen out the velocity oscillations of the fingers.

PR #177, opened 2021-10-08
fix: adds gazebo gravity vector to the gravity calculation

rickstaa:
This pull request makes sure that the gravity calculation uses the gravity vector that comes from the Gazebo simulation. 

In the old version the default value set in the [model_base.h](https://github.com/frankaemika/franka_ros/blob/1922fc9d24e37b2a8226c522e736eaa9a65d66ca/franka_hw/include/franka_hw/model_base.h) class was used (i.e. `[0, 0, -9.81]`) while gazebo uses the following gravity vector `[0, 0, -9.8000000000000007]`.

https://github.com/frankaemika/franka_ros/blob/1922fc9d24e37b2a8226c522e736eaa9a65d66ca/franka_hw/include/franka_hw/model_base.h#L236-L238

This is because the `gravity_earth` argument is not used.

https://github.com/frankaemika/franka_ros/blob/1922fc9d24e37b2a8226c522e736eaa9a65d66ca/franka_gazebo/src/franka_hw_sim.cpp#L306

### Other todos
- [ ] Update documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 * **BREAKING** Make `/panda` namespace of `franka_gazebo` optional
 * Add effort joint trajectory controller to be used by MoveIT
 * Make finger collisions primitive in `franka_gazebo`
+* add 'gravity_vector' gravity ROS parameter to FrankaHWSim
 
 ## 0.8.1 - 2021-09-08
 
--- a/franka_gazebo/src/franka_hw_sim.cpp
+++ b/franka_gazebo/src/franka_hw_sim.cpp
@@ -41,6 +41,11 @@ bool FrankaHWSim::initSim(const std::string& robot_namespace,
 
   ROS_INFO_STREAM_NAMED("franka_hw_sim", "Using physics type " << physics->GetType());
 
+  // Retrieve initial gravity vector from Gazebo
+  // NOTE: Can be overwritten by the user via the 'gravity_vector' ROS parameter.
+  auto gravity = physics->World()->Gravity();
+  this->gravity_earth_ = {gravity.X(), gravity.Y(), gravity.Z()};
+
   // Generate a list of franka_gazebo::Joint to store all relevant information
   for (const auto& transmission : transmissions) {
     if (transmission.type_ != "transmission_interface/SimpleTransmission") {
@@ -200,7 +205,7 @@ void FrankaHWSim::initFrankaModelHandle(
         " joints were found beneath the <transmission> tag, but 2 are required.");
   }
 
-  for (auto& joint : transmission.joints_) {
+  for (const auto& joint : transmission.joints_) {
     if (not urdf.getJoint(joint.name_)) {
       if (not urdf.getJoint(joint.name_)) {
         throw std::invalid_argument("Cannot create franka_hw/FrankaModelInterface for robot '" +
@@ -307,7 +312,7 @@ void FrankaHWSim::readSim(ros::Time time, ros::Duration period) {
 }
 
 void FrankaHWSim::writeSim(ros::Time /*time*/, ros::Duration /*period*/) {
-  auto g = this->model_->gravity(this->robot_state_);
+  auto g = this->model_->gravity(this->robot_state_, this->gravity_earth_);
 
   for (auto& pair : this->joints_) {
     auto joint = pair.second;
@@ -352,6 +357,11 @@ bool FrankaHWSim::readParameters(const ros::NodeHandle& nh, const urdf::Model& u
     std::string EE_T_K;  // NOLINT [readability-identifier-naming]
     nh.param<std::string>("EE_T_K", EE_T_K, "1 0 0 0 0 1 0 0 0 0 1 0 0 0 0 1");
     this->robot_state_.EE_T_K = readArray<16>(EE_T_K, "EE_T_K");
+
+    std::string gravity_vector;
+    if (nh.getParam("gravity_vector", gravity_vector)) {
+      this->gravity_earth_ = readArray<3>(gravity_vector, "gravity_vector");
+    }
 
     // Only nominal cases supported for now
     std::vector<double> lower_torque_thresholds = franka_hw::FrankaHW::getCollisionThresholds(
--- a/franka_gazebo/include/franka_gazebo/franka_hw_sim.h
+++ b/franka_gazebo/include/franka_gazebo/franka_hw_sim.h
@@ -90,6 +90,8 @@ class FrankaHWSim : public gazebo_ros_control::RobotHWSim {
   void eStopActive(const bool active) override;
 
  private:
+  std::array<double, 3> gravity_earth_;
+
   std::string arm_id_;
   gazebo::physics::ModelPtr robot_;
   std::map<std::string, std::shared_ptr<franka_gazebo::Joint>> joints_;


franka_gazebo/src/franka_hw_sim.cpp
```diff
@@ -41,6 +41,10 @@ bool FrankaHWSim::initSim(const std::string& robot_namespace,
 
   ROS_INFO_STREAM_NAMED("franka_hw_sim", "Using physics type " << physics->GetType());
 
+  // Retrieve physics parameters
+  auto gravity_ = physics->World()->Gravity();
```
gollth: This should be named simply `gravity` without the trailing underscore according to the [clang-tidy rules for variable naming](https://github.com/frankaemika/franka_ros/blob/develop/.clang-tidy#L21-L22).

Don't know why the CI didn't complain, will have a look into that

gollth (commented):
Thank you for this improvement @rickstaa, definitely makes sense!

I could imagine use cases, though, where you want to provide custom directions/strengths for the gravity vector. For example one might want to simulate their controller if the robot is in free space. For that the gravity in Gazebo would be set to zero. The Master Controller of the real robot always compensates the gravity by ~9.81 m/s^2, which the controller of such user would need to account for. 

Other use cases from the top of my head:
* Simulation of the Robot mounted non-tabletop, say on a wall
* Simulation of the robot on different places with different strengths of gravity vectors (poles, equator...)

To encompass these use cases as well, we need a way to specify if `franka_hw_sim` should use the gravity vector from Gazebo or a custom one.

My suggestion would be to make this configurable by ROS parameter. When the parameter is set, then this would be taken as gravity vector. If it is missing, then we could read it from Gazebo. This would also make a good default IMO. 

If you agree with this approach, I would like to ask you to implement this. For reference you can have a look at `FrankaHWSim::readParameter()` (probably the place, where this logic belongs as well):

https://github.com/frankaemika/franka_ros/blob/develop/franka_gazebo/src/franka_hw_sim.cpp#L334-L380

franka_gazebo/src/franka_hw_sim.cpp
gollth: Also please add a sentence in the CHANGELOG

rickstaa: @gollth Done. I also rebased onto the `develop` branch.  Please let me know when I need to change something else. 

Please note that this pull request only reduces the drift but does not remove it.

rickstaa:
> Thank you for this improvement @rickstaa, definitely makes sense!
> 
> I could imagine use cases, though, where you want to provide custom directions/strengths for the gravity vector. For example one might want to simulate their controller if the robot is in free space. For that the gravity in Gazebo would be set to zero. The Master Controller of the real robot always compensates the gravity by ~9.81 m/s^2, which the controller of such user would need to account for.
> 
> Other use cases from the top of my head:
> 
>     * Simulation of the Robot mounted non-tabletop, say on a wall
> 
>     * Simulation of the robot on different places with different strengths of gravity vectors (poles, equator...)
> 
> 
> To encompass these use cases as well, we need a way to specify if `franka_hw_sim` should use the gravity vector from Gazebo or a custom one.
> 
> My suggestion would be to make this configurable by ROS parameter. When the parameter is set, then this would be taken as gravity vector. If it is missing, then we could read it from Gazebo. This would also make a good default IMO.
> 
> If you agree with this approach, I would like to ask you to implement this. For reference, you can have a look at `FrankaHWSim::readParameter()` (probably the place, where this logic belongs as well):
> 
> https://github.com/frankaemika/franka_ros/blob/develop/franka_gazebo/src/franka_hw_sim.cpp#L334-L380

I like your suggestion. I will add it to this pull request.

CHANGELOG.md
```diff
@@ -2,11 +2,13 @@
 
 ## 0.x - UNRELEASED
 
+
```
gollth: Please keep whitespace changes to a minimum to avoid conflicts

rickstaa: @gollth Sorry I overlooked that one. I will remove it in the next force push.

rickstaa: Fixed I will now apply your suggestions to the pull request.

franka_gazebo/src/franka_hw_sim.cpp
```diff
@@ -331,7 +336,7 @@ void FrankaHWSim::writeSim(ros::Time /*time*/, ros::Duration /*period*/) {
 
 void FrankaHWSim::eStopActive(bool /* active */) {}
 
-bool FrankaHWSim::readParameters(const ros::NodeHandle& nh, const urdf::Model& urdf) {
+auto FrankaHWSim::readParameters(const ros::NodeHandle& nh, const urdf::Model& urdf) -> bool {
```
rickstaa: Automatically applied by clang-tidy.

franka_gazebo/src/franka_hw_sim.cpp
```diff
@@ -17,11 +17,11 @@
 
 namespace franka_gazebo {
 
-bool FrankaHWSim::initSim(const std::string& robot_namespace,
+auto FrankaHWSim::initSim(const std::string& robot_namespace,
                           ros::NodeHandle model_nh,
                           gazebo::physics::ModelPtr parent,
                           const urdf::Model* const urdf,
-                          std::vector<transmission_interface::TransmissionInfo> transmissions) {
+                          std::vector<transmission_interface::TransmissionInfo> transmissions) -> bool {
```
rickstaa: Automatically applied by clang-tidy. I can remove the automatic clang-tidy changes if you want.

franka_gazebo/src/franka_hw_sim.cpp
```diff
@@ -200,7 +205,7 @@ void FrankaHWSim::initFrankaModelHandle(
         " joints were found beneath the <transmission> tag, but 2 are required.");
   }
 
-  for (auto& joint : transmission.joints_) {
+  for (const auto & joint : transmission.joints_) {
```
rickstaa: Automatically applied by clang-tidy.

rickstaa:
@gollth, @marcbone I implemented the requested changes and rebased on the `develop` branch. I used `g_vector` for the gravity ROS parameter, but I can change it to your liking.

franka_gazebo/src/franka_hw_sim.cpp
marcbone: the const in the for loop can stay, however the trailing return types, look just look ugly imo. We currently use clang-tidy 7 in the CI. There are lots of things that could be fixed with a newer version of clang-tidy, however there are other lints like  modernize-use-trailing-return-type or readability-function-cognitive-complexity that you do not want to have. Therefore, we currently stick to clang-tidy 7.

franka_gazebo/src/franka_hw_sim.cpp
marcbone: There is a space between the auto and the &. Actually this should not pass the CI :thinking:

rickstaa: Ah, that's strange.

I also noticed that the [clang-tidy vscode](https://marketplace.visualstudio.com/items?itemName=notskm.clang-tidy) doesn't take that rule into account when formatting. This can of course be a bug with the extensions and not the `.clang-tidy` config. If it is a bug in the extension it will probably be fixed when the C++ extension adds clang-tidy support (see https://github.com/microsoft/vscode-cpptools/issues/2908).

franka_gazebo/src/franka_hw_sim.cpp
rickstaa: Strange that the extension does not pick up the right clang-tidy config. I reverted the clang-tidy changes manually.

gollth:
I vote for `gravity_vector` for clarity, other than that, LGTM

rickstaa:
Great I updated the code to reflect that change.

franka_gazebo/src/franka_hw_sim.cpp
marcbone: It is because we include modernize* and readability*, so if new checks appear they will be included by default. Thats why you need to choose the correct clang-tidy version, so just use clang-tidy 7.

rickstaa: Ah over overlooked that in your previous answer. I'm on clang-tidy 10.

gollth (approved):

rickstaa:
Thanks a lot for merging this!

PR #147, opened 2021-08-13
Improve gazebo hand/finger collision behavoir

rickstaa:
This pull request improves the Gazebo collision geometries.

#### Old collision geometries

image
image

#### New collision geometries

image
image

Issue #179, opened 2021-10-18
[bug] Pull request #147 breaks `gripper_action` actions


rickstaa:
@falfab I just noticed that my previous pull request breaks the `/panda/franka_gripper/gripper_action/goal` action (see 1 gif below). I'm now investigating why this is the case, as this did not happen in the [justagist/franka_panda_description](https://github.com/justagist/franka_panda_description) (i.e. the repository from which the `stl` files are taken). This problem does not occur when using the `/panda/franka_gripper/move/action` here, however the gripper is not able to close fully (see gif 2 below).

![gripper_problem](https://user-images.githubusercontent.com/17570430/137703335-43d8611b-eeda-4797-a6b1-bbdafdb19ff8.gif)

![gripper_problem_move](https://user-images.githubusercontent.com/17570430/137703735-086f8bd8-b67d-4c44-ab58-ec6555a06479.gif)

I spend some minutes trying to improve the mesh but did not find one for which the problems were solved. Although I think improving the collision mesh can remove the problems, I do not have more time to work on this. As a workaround I created #180, which contains more accurate collision geometries, to solve the problems.

#### Removed top part intrusion

When we remove the top part intrusion, the behaviour is not improved.

[finger_improved.zip](https://github.com/frankaemika/franka_ros/files/7364641/finger_improved.zip)
 
![image](https://user-images.githubusercontent.com/17570430/137727314-2150c92d-b186-4f34-a6bb-2c6ff407d7e7.png)

#### Shorten top

When we shorten the top part, the behaviour is not improved.

[finger_improved3.zip](https://github.com/frankaemika/franka_ros/files/7364720/finger_improved3.zip)

![image](https://user-images.githubusercontent.com/17570430/137729128-79c3db93-916d-4810-83aa-0608b47b5c62.png)

#### Lengthen gripper pads

[finger_improved4.zip](https://github.com/frankaemika/franka_ros/files/7364840/finger_improved4.zip)

When I lengthen the bottom (i.e., grasping pads) by 1 mm the gripper now spawns in the closed position. When closing the gripper still strange behaviour is present.

rickstaa:
This was fixed by 936ef8ed5d7eee7499fe5a11d6233f84e3d4eaf8.

PR #180, opened 2021-10-18
Fixes gazebo gripper action bug

rickstaa:
This pull request fixes a bug that was introduced by #147. In #147 the finger collision geometries were replaced by more accurate meshes. These more accurate messages however broke the `gripper_action` action. This commit migrates back to using geometries for the finger collision geometries. The action server works fine when geometries are used.

### DIFF
--- a/franka_description/robots/panda_gazebo.xacro
+++ b/franka_description/robots/panda_gazebo.xacro
@@ -309,8 +309,21 @@
         </geometry>
       </visual>
       <collision>
+        <origin xyz="0 0.0184 0.0101" rpy="0 0 0"/>
         <geometry>
-          <mesh filename="package://franka_description/meshes/collision/finger.stl" />
+          <box size="0.022 0.014 0.0177" />
+        </geometry>
+      </collision>
+      <collision>
+        <origin xyz="0 0.01655 0.02755" rpy="${3.257/(2*pi)} 0 0"/>
+        <geometry>
+          <box size="0.022 0.00686 0.0237" />
+        </geometry>
+      </collision>
+      <collision>
+        <origin xyz="0 0.0073 0.045" rpy="0 0 0"/>
+        <geometry>
+          <box size="0.022 0.0147 0.0178" />
         </geometry>
       </collision>
     </link>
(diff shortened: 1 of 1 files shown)


rhaschke:
I don't see how the gripper action is affected by the collision geometries. What exactly is broken?

rickstaa:
Some time ago, I created #147 to make the finger collision geometries more accurate in the gazebo simulation. This pull request was merged into the `develop` branch. Unfortunately, changing to meshes causes strange behaviours in the gazebo simulation that I do not fully understand (see the section below and #179). #180 is a hotfix to temporary fix these behaviours.

#### Problem description

Multiple problems appeared when I changed to meshes. First, the gripper is now not fully closed when the gazebo simulation is started:

image

More problematic the `franka_gripper/gripper_action` now is unstable:

**with geometries (the hotfix in this pull request)**

with geometries

**with meshes**

with meshes

This is caused by the fact the the `gripper_action` transitions to the `HOLDING` state when the position is reached:

https://github.com/frankaemika/franka_ros/blob/da13eb1d6f5cc9600dbb22a60759016e715e8d27/franka_gazebo/src/franka_gripper_sim.cpp#L435-L439

https://github.com/frankaemika/franka_ros/blob/da13eb1d6f5cc9600dbb22a60759016e715e8d27/franka_gazebo/src/franka_gripper_sim.cpp#L183-L190

https://github.com/frankaemika/franka_ros/blob/da13eb1d6f5cc9600dbb22a60759016e715e8d27/franka_gazebo/src/franka_gripper_sim.cpp#L135-L142

It looks that with these meshes, something goes wrong in the interaction between the finger meshes and the hand when the fingers keep applying a force. This can be seen from the fact that the behaviour is not present in the `franka_gripper:move` action (which only moves to the position):

with meshes move

Did you encounter this behaviour before? This issue is closely related to #173.

rickstaa:
@gollth I verified 936ef8ed5d7eee7499fe5a11d6233f84e3d4eaf8 and the problems are gone. The simulated gripper now works perfectly with the `franka_gripper/move` action. I however noticed that the `franka_gripper/gripper_action` still doesn't give the desired behaviour (see gif below and https://github.com/frankaemika/franka_ros/pull/173#issuecomment-966255718). I will therefore keep #172 open but close #180 and #179. 

gripper_action_bug

I think the behaviour that was pointed out by @rhaschke in https://github.com/frankaemika/franka_ros/pull/173#issuecomment-966255718 would be the desired behaviour.

> IMO, the gripper action should realize position-control as long as measured efforts are smaller than max_effort. Otherwise, it should switch to force-control and maintain the given (grasp) force.

PR #183, opened 2021-10-25
Cosmetic changes to xacro and launch files

rhaschke:
This PR comprises some simple cosmetic changes and cleanup to xacro files.
Please don't squash-merge. All commits are rather independent. See individual commit messages for details.

PR #187, opened 2021-10-27
Enabling non-namespace use of Gazebo robot

rhaschke:
So far, the Gazebo robot was always started in the `arm_id=panda` namespace. However, the real robot is not running in a namespace. To increase coherence, this PR (**building on top of #183** to avoid merge conflicts) adds two commits:
- introducing the `ns` argument to `panda_arm.urdf.xacro`
- refactoring `franka_gazebo/panda.launch` to allow an empty namespace

This essentially decouples the `namespace` from the `arm_id` and - for example - allows spawning two arms as follows:
```bash
roslaunch gazebo_ros empty_world.launch world_name:=worlds/empty.world use_sim_time:=true
roslaunch franka_gazebo panda.launch gazebo:=false arm_id:=panda1 ns:=left y:=0
roslaunch franka_gazebo panda.launch gazebo:=false arm_id:=panda2 ns:=right y:=1
```

### DIFF
--- a/franka_description/robots/dual_panda_example.urdf.xacro
+++ b/franka_description/robots/dual_panda_example.urdf.xacro
@@ -26,11 +26,11 @@
   </link>
 
   <!-- right arm with gripper -->
-  <xacro:panda_arm arm_id="$(arg arm_id_1)" connected_to="base"  xyz="0 -0.5 1" safety_distance="0.03"/>
+  <xacro:panda_arm arm_id="$(arg arm_id_1)" ns="$(arg arm_id_1)" connected_to="base" xyz="0 -0.5 1" safety_distance="0.03" />
   <xacro:hand ns="$(arg arm_id_1)" rpy="0 0 ${-pi/4}" connected_to="$(arg arm_id_1)_link8" safety_distance="0.03"/>
 
   <!-- left arm with gripper -->
-  <xacro:panda_arm arm_id="$(arg arm_id_2)" connected_to="base"  xyz="0 0.5 1" safety_distance="0.03"/>
+  <xacro:panda_arm arm_id="$(arg arm_id_2)" ns="$(arg arm_id_2)" connected_to="base" xyz="0 0.5 1" safety_distance="0.03" />
   <xacro:hand ns="$(arg arm_id_2)" rpy="0 0 ${-pi/4}" connected_to="$(arg arm_id_2)_link8" safety_distance="0.03"/>
 
 </robot>
--- a/franka_description/robots/panda_arm.urdf.xacro
+++ b/franka_description/robots/panda_arm.urdf.xacro
@@ -7,6 +7,8 @@
   <xacro:arg name="hand" default="false" />
   <!-- Is the robot being simulated in gazebo?" -->
   <xacro:arg name="gazebo" default="false" />
+  <!-- Namespace of robot in Gazebo/ROS (default: empty) -->
+  <xacro:arg name="ns" default="" />
 
   <xacro:unless value="$(arg gazebo)">
     <!-- Create a URDF for a real hardware -->
@@ -61,7 +63,7 @@
 
     <gazebo>
       <plugin name="gazebo_ros_control" filename="libgazebo_ros_control.so">
-        <robotNamespace>$(arg arm_id)</robotNamespace>
+        <robotNamespace>$(arg ns)</robotNamespace>
         <controlPeriod>0.001</controlPeriod>
         <robotSimType>franka_gazebo/FrankaHWSim</robotSimType>
       </plugin>
--- a/franka_gazebo/launch/panda.launch
+++ b/franka_gazebo/launch/panda.launch
@@ -43,6 +43,7 @@
                     gazebo:=true
                     hand:=$(arg use_gripper)
                     arm_id:=$(arg arm_id)
+                    ns:=$(arg arm_id)
                     xyz:='$(arg x) $(arg y) $(arg z)'
                     rpy:='$(arg roll) $(arg pitch) $(arg yaw)'">
     </param>
--- a/franka_gazebo/launch/panda.launch
+++ b/franka_gazebo/launch/panda.launch
@@ -1,5 +1,6 @@
 <?xml version="1.0"?>
 <launch>
+  <arg name="ns" default="" doc="Namespace to launch the Gazebo robot in" />
 
   <!-- Gazebo & GUI Configuration -->
   <arg name="gazebo"      default="true"  doc="Should the gazebo simulation be launched? Use false in case if you want to include this file and launch gazebo yourself" />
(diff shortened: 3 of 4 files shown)


rhaschke:
Note that using a `namespace` different than `arm_id` triggers this warning: 
https://github.com/frankaemika/franka_ros/blob/563465d92325d7005085b0417e85678eadaec2f4/franka_gazebo/src/franka_hw_sim.cpp#L26-L33

... which was introduced only recently by @gollth in f2f82b2e4c3279d7196d1836049d649bd4252fb3. I think this commit already resolved the main problems to separate the name of the robot (`arm_id`) from the ROS namespace it is running in. In other words, I think this warning is superfluous.

rhaschke:
Rebased after merge of #183.

gollth:
Hmm, not sure about this one. 

I think the main confusion here is the distinction between *topic namespaces* and URDF prefixes, which are two distinct features. Let's try to not mix the two. 

Passing a topic namespace into an URDF seems already flawed to me. Ideally a URDF does not need to care in which namespace it is loaded. It is valid in any namespace.

The only reason I see where we would need the namespace in the URDF anyways would be for 

```xml
<gazebo>
  <plugin name="gazebo_ros_control" filename="libgazebo_ros_control.so">
    <robotNamespace>$(arg arm_id)</robotNamespace> <!-- HERE -->
    <controlPeriod>0.001</controlPeriod>
    <robotSimType>franka_gazebo/FrankaHWSim</robotSimType>
  </plugin>
</gazebo>
```

The [documentation for this](http://gazebosim.org/tutorials/?tut=ros_control) `<robotNamespace>` states:

> The ROS namespace to be used for this instance of the plugin, defaults to robot name in URDF/SDF

So can we just leave it out completely? Then we don't have any association between ROS topic namespaces in the URDF anymore:

```xml
<gazebo>
  <plugin name="gazebo_ros_control" filename="libgazebo_ros_control.so">
    <controlPeriod>0.001</controlPeriod>
    <robotSimType>franka_gazebo/FrankaHWSim</robotSimType>
  </plugin>
</gazebo>
```

Regarding the naming between `franka_gazebo` and `franka_control` I think you have a fair point, that the two should match. When we let gazebo figure out the namespace by itself, we can even remove the `<group ns="$(arg arm_id)">` entirely from the `panda.launch` file. This has two advantages:

* No special logic for dealing with empty namespaces like you suggested
* Give user code (launch files) the possibility to declare their namespaces, e.g.:

```xml
<launch>
  <group ns="panda_left">
    <include file="$(find franka_gazebo)/launch/panda.launch">
      <arg name="arm_id"     value="panda_1" />
    </include>
  </group>

  <group ns="panda_right">
    <include file="$(find franka_gazebo)/launch/panda.launch">
      <arg name="arm_id"     value="panda_2" />
    </include>
  </group>
</include>
```

Notice how the namespace could potentially even be different than the `arm_id`

I drafted a possible solution [here](https://github.com/frankaemika/franka_ros/commit/f9ea5b15924b262c9e6e7ef9731c9452272eeb2c) if you want to have a look. Would that be compatible with MoveIT then?

---

> This essentially decouples the `namespace` from the `arm_id` and - for example - allows spawning two arms as follows:

BTW. this is already possible, see https://frankaemika.github.io/docs/franka_ros.html#customization

gollth:
> Note that using a `namespace` different than `arm_id` triggers this warning:

Regarding this check I also agree, that, in any case this will be outdated then. The main idea behind this was to give the User a clearer warning if the namespaces don't match other than the cryptic `Controller Spawner couldn't find the expected controller_manager ROS interface.` when the URDF could not be loaded.

However now I see, that if this warning would be printed, then the URDF had to be found in the first place to define the `FrankaHWSim` plugin...

rhaschke:
> I think the main confusion here is the distinction between _topic namespaces_ and URDF prefixes, which are two distinct features. Let's try to not mix the two.

Good that we agree on this crucial point. The problem is that the old code _linked_ these two features, using the URDF prefix `arm_id` _also_ for the topic namespaces generated by Gazebo:
https://github.com/frankaemika/franka_ros/blob/da8a709fd69595fd68df7b5979eafdb837e78295/franka_description/robots/panda_arm.urdf.xacro#L64-L65

Thus, the central contribution of this PR is to [change this line](https://github.com/frankaemika/franka_ros/pull/187/files#diff-ee41246bb15e7465711a7e8986dfcabbb79ac3d7520803a4052292e7b4667f1eL64). The rest is required to forward the `ns` argument to that spot.
I like your idea to "let Gazebo figure out the namespace by itself". However, [your draft](https://github.com/frankaemika/franka_ros/commit/f9ea5b15924b262c9e6e7ef9731c9452272eeb2c) doesn't achieve that: Your solution relies on the fact that Gazebo itself is started within the corresponding namespace. However, this rules out the possibility to spawn multiple robots with different arm IDs _and_ namespaces into the _same_ Gazebo instance.
I don't see an option other than specifying the Gazebo namespace in the URDF. One could try to use the namespace of the [robot spawner](https://github.com/frankaemika/franka_ros/commit/f9ea5b15924b262c9e6e7ef9731c9452272eeb2c), but this would require changes to `gazebo_ros_control`.

gollth:
Don't know if I understand your point entirely. 

> However, [your draft](https://github.com/frankaemika/franka_ros/commit/f9ea5b15924b262c9e6e7ef9731c9452272eeb2c) doesn't achieve that

For me it does: If I start this launchfile with the changes from the above commit, I can start both pandas in separate namespaces (which can even be different from the `arm_id`s) and Gazebo in its own (root) namespace:

```xml
<?xml version="1.0"?>
<launch>

  <include file="$(find gazebo_ros)/launch/empty_world.launch" >
    <!-- Start paused, simulation will be started, when Pandas were loaded -->
    <arg name="paused" value="true"/>
    <arg name="use_sim_time" value="true"/>
  </include>

  <group ns="panda_left">
    <include file="$(find franka_gazebo)/launch/panda.launch">
      <arg name="arm_id"     value="panda_1" />
      <arg name="y"          value="-0.5" />
      <arg name="gazebo"     value="false" />
      <arg name="paused"     value="true" />
    </include>
  </group>

  <group ns="panda_right">
    <include file="$(find franka_gazebo)/launch/panda.launch">
      <arg name="arm_id"     value="panda_2" />
      <arg name="y"          value="0.5" />
      <arg name="gazebo"     value="false" />
      <arg name="paused"     value="false" />
    </include>
  </group>

</launch>
```

Output of `rosnode list` confirms that:

```
/gazebo
/gazebo_gui
/panda_left/interactive_marker
/panda_left/joint_state_desired_publisher
/panda_left/joint_state_publisher
/panda_left/panda_1_controller_spawner
/panda_left/panda_1_gripper_spawner
/panda_left/robot_state_publisher
/panda_right/joint_state_desired_publisher
/panda_right/joint_state_publisher
/panda_right/panda_2_controller_spawner
/panda_right/panda_2_gripper_spawner
/panda_right/robot_state_publisher
/rosout
```

Can you elaborate what you mean by "Your solution relies on the fact that Gazebo itself is started within the corresponding namespace"?

rhaschke:
You are totally right. I just verified that the Gazebo spawner indeed considers the namespace it is started in. My statement in https://github.com/frankaemika/franka_ros/pull/187#issuecomment-963293015 was based on the (unverified - sorry, I was in a rush) assumption that this wouldn't be the case. Your solution is much better then. Closing here in favour of #196.

gollth:
No worries, thanks for the contribution

rickstaa:
> Note that using a `namespace` different than `arm_id` triggers this warning:
> 
> https://github.com/frankaemika/franka_ros/blob/563465d92325d7005085b0417e85678eadaec2f4/franka_gazebo/src/franka_hw_sim.cpp#L26-L33
> 
> ... which was introduced only recently by @gollth in [f2f82b2](https://github.com/frankaemika/franka_ros/commit/f2f82b2e4c3279d7196d1836049d649bd4252fb3). I think this commit already resolved the main problems to separate the name of the robot (`arm_id`) from the ROS namespace it is running in. In other words, I think this warning is superfluous.



> However now I see, that if this warning would be printed, then the URDF had to be found in the first place to define the `FrankaHWSim` plugin...

@gollth, @Maverobot I created a pull request that removes the redundant ROS namespace warning from FrankaHWSim. Please feel free to ask for changes if you think of a better solution.

PR #189, opened 2021-11-02
Rework hand collision

rhaschke:
This is part of implementing https://github.com/ros-planning/moveit/issues/2936, i.e. providing two separate links for detailed and coarse hand collision:
- `panda_hand` uses the detailed .stl mesh to enable collision checking with the environment
- `panda_hand_coarse` uses coarse geometries to mimic self-collision checking of the internal controller

This goes in line with https://github.com/ros-planning/panda_moveit_config/pull/95.

### DIFF
--- a/franka_description/robots/hand.xacro
+++ b/franka_description/robots/hand.xacro
@@ -12,9 +12,16 @@
     <link name="${ns}_hand">
       <visual>
         <geometry>
-          <mesh filename="package://franka_description/meshes/visual/hand.dae"/>
+          <mesh filename="package://franka_description/meshes/visual/hand.dae" />
         </geometry>
       </visual>
+      <collision>
+        <geometry>
+          <mesh filename="package://franka_description/meshes/collision/hand.stl" />
+        </geometry>
+      </collision>
+    </link>
+    <link name="${ns}_hand_coarse">
       <collision>
         <origin xyz="0 0 0.04" rpy="0 ${pi/2} ${pi/2}"/>
         <geometry>
@@ -52,6 +59,10 @@
         </geometry>
       </collision>
     </link>
+    <joint name="${ns}_hand_coarse_joint" type="fixed">
+      <parent link="${ns}_hand" />
+      <child link="${ns}_hand_coarse" />
+    </joint>
     <link name="${ns}_leftfinger">
       <visual>
         <geometry>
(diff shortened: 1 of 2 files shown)


rickstaa (approved):

rhaschke:
For _environmental_ collisions, the same problem arises as for the hand: While planning might fail with the coarse self-collision models, the robot might perfectly fit into a narrow passage with the detailed meshes. 
Thus, this PR should be extended to apply the same idea to the whole arm, i.e. define detailed meshes for environmental collisions and use separate coarse models for self-collision checking only. 
Instead of the name `coarse`, we should use another suffix, e.g. `sc` to indicate their use for self-collision checking. 

When reworking the URDF, I also suggest merging [`panda_gazebo.xacro`](https://github.com/frankaemika/franka_ros/blob/develop/franka_description/robots/panda_gazebo.xacro) into [`panda_arm.xacro`](https://github.com/frankaemika/franka_ros/blob/develop/franka_description/robots/panda_arm.xacrol) and [`hand.xacro`](https://github.com/frankaemika/franka_ros/blob/develop/franka_description/robots/hand.xacro). The Gazebo-specific file duplicates much of the structure of the basic URDF and then just adds inertial properties. 

My suggestion is to factor out the inertial properties into a yaml dictionary:
```yaml
link0:
  origin:
    xyz: 0 0 0
    rpy: 0 0 0
  mass:
  inertia:
    xx: 0
    yy: 0
    zz: 0
    xy: 0
    xz: 0
    yz: 0
link 1:
  ...
```
and use them in xacro essentially like this:
```xml
        <inertial>
          <origin rpy="${inertial.origin.rpy}" xyz="${inertial.origin.xyz}" />
          <mass value="${inertial.mass}" />
          <xacro:property name="I" value="${inertial.inertia}" />
          <inertia ixx="${I.xx}" ixy="${I.xy}" ixz="${I.xz}"
                                 iyy="${I.yy}" iyz="${I.yz}"
                                               izz="${I.zz}" />
        </inertial>
```

So, there is more work todo! As I need to focus on other stuff again, I cannot contribute very much to this PR in the near future. So help is highly welcome. I drafted a basic proof-of-concept of this proposal in #199.

rhaschke:
Closing in favour of #199.

PR #192, opened 2021-11-07
Remap joint states instead of re-publish

rhaschke:
Re-publishing `joint_states` via `joint_state_publisher` (a python app) is highly inefficient. 
IMO remapping would be the right way to go instead - as long nobody needs the published topics on the namespace 
`franka_state_controller` / `franka_gripper`. 

But, remapping has its drawbacks too:
- It breaks existing software relying on topics `franka_state_controller/joint_states` and/or `franka_gripper/joint_states`
- `joint_states` messages from arm and gripper are published both to the same topic, so we cannot easily distinguish them anymore (w/o looking at the message of course)
- The variable message format (7 vs. 2 joints) might confuse existing software, e.g. `rqt_plot`.

Note: This PR includes #191.

### DIFF
--- a/franka_control/launch/franka_control.launch
+++ b/franka_control/launch/franka_control.launch
@@ -22,10 +22,4 @@
     <rosparam unless="$(arg load_gripper)" param="source_list">[franka_state_controller/joint_states] </rosparam>
     <param name="rate" value="30"/>
   </node>
-  <node name="joint_state_desired_publisher" type="joint_state_publisher" pkg="joint_state_publisher" output="screen">
-    <rosparam if="$(arg load_gripper)" param="source_list">[franka_state_controller/joint_states_desired, franka_gripper/joint_states] </rosparam>
-    <rosparam unless="$(arg load_gripper)" param="source_list">[franka_state_controller/joint_states_desired] </rosparam>
-    <param name="rate" value="30"/>
-    <remap from="/joint_states" to="/joint_states_desired" />
-  </node>
 </launch>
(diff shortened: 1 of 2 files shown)


rhaschke:
@frankaemika, what do you think of this idea? Nobody ever commented on this for 9 months.

rickstaa:
@rhaschke, thank you for your work on improving the `franka_ros` package's performance! I greatly appreciate the changes made in this pull request and tried to review it to speed up the merge process.

During my review, I encountered some issues that I believe are important to address:

1. **Unrecognized 'remap' Tags:** When I attempted to run `roslaunch franka_gazebo panda.launch`, I received warnings about unrecognized 'remap' tags within the `<include>` tag.

2. **Simulation Breakage:** The simulation is broken as no messages are being published on the `/joint_states` topic. To temporarily address this, I moved the 'remap' command outside of the `<include>` tag. However, this results in two separate messages being published sequentially—one for arm joint states and another for gripper joint states. 

Considering that using the `joint_state_publisher` is sub-optimal, perhaps we could explore a custom C++ joint publisher solution. This would involve merging arm and gripper joint states and republishing them. We might take inspiration from tools like [relay.cpp](https://github.com/ros/ros_comm/blob/noetic-devel/tools/topic_tools/src/relay.cpp) in the ROS repository.

@gollth, @@FE-EnricoSartori I think @rhaschke raised a vital performance improvement opportunity. I'm interested to know what you think about this change 🤔.

PR #195, opened 2021-11-08
fix interactive marker initialization

JohannesPankert:
Wait until an initial pose is found unless rospy is in shutdown state.


marcbone:
Thank you for your contribution. We have already fixed this internally. I have just pushed it to the develop branch. Therefore, I am going to close this. I am sorry for the inconveniences that this may has caused you.

Issue #193, opened 2021-11-07
Requirements for panda_moveit_config release


rhaschke:
The new release of https://github.com/ros-planning/panda_moveit_config is blocked by a few things in `franka_ros`:
- [x] #199
- [x] #183
- [x] #196
- [x] ~~#190~~
  Covered by #199
- [x] ~~#186~~ -> e899240365797de6d53c222d6512b74c3a40268e
- [x] #194
- [x] ~~#173~~

I have merged these (and the first item from the nice-to-have list) into #188 for convenience.

### Nice to have
- [x] #191
- [x] Allow to use different hardware transmissions in Gazebo: https://github.com/frankaemika/franka_ros/commit/2514ae1990d51ca808e6c49d22d24ae7a99db532

### Optional
- [ ] #192

### Open issues
- [ ] GripperCommand doesn't signal convergence (but movements are correctly executed, see [#173](https://github.com/frankaemika/franka_ros/pull/173#issuecomment-933435669))
  - MoveIt controller manager fails with: `Controller is taking too long to execute trajectory`
  - `/franka_gripper/gripper_action/status` reports: 
    `Unexpected state transistion: The gripper not in HOLDING as expected`
- [x] #179
- [ ] #172

rickstaa:
@rhaschke Thanks for collecting all our issues into an overview. One small suggestion can you change

> GripperCommand doesn't signal convergence (but movements are correctly executed) 

to 

> GripperCommand doesn't signal convergence (but movements are correctly executed) (see #173)

The problem is explained in detail in that pull request. Further, another related issue that should be mentioned is that #146, which was merged to the devel branch, broke the gripper actions (see #179).

gollth:
For the record, @rhaschke shall we change PR #173 to Issue [#172](https://github.com/frankaemika/franka_ros/issues/172) on the checklist? (as discussed [here](https://github.com/frankaemika/franka_ros/pull/173#issuecomment-965023144))

rickstaa:
>A few things block the new release of https://github.com/ros-planning/panda_moveit_configrankaemika/franka_ros/pull/191)
> * [ ]  Allow to use different hardware transmissions in Gazebo than the default effort transmission
>   
>   * [ ]  [Adds the 'transmission' arg to the 'panda_arm.urdf.xacro' file #157](https://github.com/frankaemika/franka_ros/pull/157)
[...]

PR #199, opened 2021-11-09
Provide detailed and coarse collision models

rhaschke:
As pointed out in https://github.com/frankaemika/franka_ros/pull/189#issuecomment-964408196, the coarse self-collision models not only pose an issue for the hand but also for the rest of the robot. However, the hand is most crucial for grasping.

This PR implements the ideas proposed in  https://github.com/frankaemika/franka_ros/pull/189#issuecomment-964408196, namely
- Defining detailed _and_ coarse collision models as proposed for the hand in #189
- Using the xacro macro `link` to reduce boilerplate code
- Merging [`panda_gazebo.xacro`](https://github.com/frankaemika/franka_ros/blob/develop/franka_description/robots/panda_gazebo.xacro) into [`panda_arm.xacro`](https://github.com/frankaemika/franka_ros/blob/develop/franka_description/robots/panda_arm.xacrol) and [`hand.xacro`](https://github.com/frankaemika/franka_ros/blob/develop/franka_description/robots/hand.xacro)
- Defining inertia properties in an external yaml file

~~Somebody from Franka (@gollth, @marcbone) or @rickstaa? should pick this up, if we agree on the general approach.~~

This PR replaces/generalizes #189.
Fixes #190 (mimic joint is part of hand.xacro).

### TODOs
- [x] The coarse collision models, which are essentially capped cylinders, should be generated using a corresponding xacro macro for capsules. I'm sure, internally Franka represents those geometries with capsules already. Thus, I leave that for you.
- [x] The virtual `link8` has some coarse collision models associated, which should better get associated to `link7`. Effectively, of course, that doesn't make a difference as both links are rigidly connected. However, semantically, it would be more appealing.
- [x] The coarse model of `link8` is composed of two very similar spheres and a cylinder that is - in my opinion - wrongly oriented. I guess, this geometry should reflect a capsule as well? Could it get replaced by a simple sphere?
  image


rickstaa:
@rhaschke Thanks a lot for Implementing this! Merging the gazebo and normal `urdf` will make franka_ros more MoveIt MSA compatible.

rhaschke:
Ping @gollth, @marcbone: I didn't get feedback yet on this restructuring attempt, which is crucial to allow more fine-grained collision checking with the environment (while keeping self-collision) as needed for grasping.

gollth:
> Ping @gollth, @marcbone: I didn't get feedback yet on this restructuring attempt, which is crucial to allow more fine-grained collision checking with the environment (while keeping self-collision) as needed for grasping.

We are still working on the other open points in https://github.com/frankaemika/franka_ros/issues/193. We will give you feedback here shortly

rhaschke:
Friendly ping, @gollth. Did you already had a chance do look into this PR? This is crucial for a new Noetic release of `panda_moveit_config` and fixing an issue introduced months ago (when introducing coarse collision models).
Meanwhile, we have merged and (just) released the required changes to MoveIt and srdfdom:
- https://github.com/ros/rosdistro/pull/31977
- https://github.com/ros/rosdistro/pull/31979

gollth:
Hi @rhaschke,

glad to hear the new release of MoveIT. We are currently quite busy with other stuff but have this collision model topic on the radar and will have a look soonish!

gollth (changes requested):
Hi @rhaschke, I will test your PR together with the new MoveIT shortly. In the meantime I can already give you some comments on the code itself.

* Please revert the `prefix` to `arm_id` to avoid breaking changes. I get your point, that here it is used as a prefix to joint names, but in all our nomenclature we use the term `arm_id`. Especially for multi-arm situation it makes sense to distinguish multiple robots by ids.
* Please revert the renaming the `panda_arm.xacro` file for backwards compatibility

franka_description/robots/arm.xacro
```diff
@@ -276,16 +256,19 @@
...
-      <child link="${arm_id}_link7"/>
+      <parent link="${prefix}link6"/>
+      <child link="${prefix}link7"/>
       <axis xyz="0 0 1"/>
       <limit effort="12" lower="-2.8973" upper="2.8973" velocity="2.6100"/>
+      <dynamics D="1" K="7000" damping="0.003" friction="0.0" mu_coulomb="0" mu_viscous="16"/>
     </joint>
-    <link name="${arm_id}_link8">
+    <link name="${prefix}link8"/>
+    <link name="${prefix}link8_sc">
```
gollth: ~~Why did you introduce a new `link8_sc` and what does `sc` stand for?~~

franka_description/robots/arm.xacro
```diff
@@ -59,21 +52,19 @@
...
+      <dynamics D="1" K="7000" damping="0.003" friction="0.0" mu_coulomb="0" mu_viscous="16"/>
     </joint>
-    <link name="${arm_id}_link2">
-      <visual>
-        <geometry>
-          <mesh filename="package://${description_pkg}/meshes/visual/link2.dae"/>
-        </geometry>
-      </visual>
+    <xacro:link name="link2">
+      <self_collision_geometries>
```
gollth: Please indent all children with all `<self_collision_geometries>` tags

franka_description/robots/inertial.yaml
```diff
@@ -0,0 +1,137 @@
+# This file does not contain official inertial properties
```
gollth, line 1: The extraction of the dynamic parameters into their own yaml file is pretty nice. I like that a lot and didn't know such a thing was possible in XACRO. However it fails in the Melodic CI. Can you have look please?

franka_description/robots/arm.xacro
gollth: ...probably self collision I guess. Now I see that your `<self_collision_geometries />` tag simply generates new links with a `_sc` suffix and the collision geometries beneath that. TBH I assumed that MoveIT would pick up the this tag and interpret it accordingly rather then looking for specifically named links...

gollth:
When I run `roslaunch panda_moveit_config franka_control.launch` on the latest `noetic-devel` branch of panda moveit config my controller manager is not able to start/find the `franka_gripper` controller:

```console
▏ERROR▕ /franka_control ⟩ ControllerManager::loadController() ⟩ L313 ⟫ Could not load controller 'franka_gripper' because controller type 'GripperCommand' does not exist.
```

Am I missing a dependency? 

@rhaschke or @rickstaa could you point me somewhere please?

franka_description/robots/arm.xacro
rhaschke: As explained in the main comment, I just introduce both, detailed and coarse collision models for each link - the latter ones only used for self-collision checking. MoveIt doesn't look for specific names. You can pick any name you want.
The semantics of the different collision models is eventually defined in the SRDF.

franka_description/robots/arm.xacro
rhaschke: I will do. I just tried to minimize the changeset.

franka_description/robots/inertial.yaml
rhaschke: I think you just use an outdated version of xacro in your Melodic docker container. 
The Dockerfiles are missing an `apt-get upgrade`.

rhaschke:
> * Please revert the `prefix` to `arm_id` to avoid breaking changes. 
I get your point, that here it is used as a prefix to joint names, but in all our nomenclature we use the term `arm_id`. Especially for multi-arm situation it makes sense to distinguish multiple robots by ids.
> * Please revert the renaming the `panda_arm.xacro` file for backwards compatibility

Note that we can maintain backwards compatibility by other means. For example, #210 illustrates how to deprecate a property in favour of a more meaningful name.
A similar approach could be used for renaming the file. I fully agree with your request to maintain backwards compatibility.
However, this shouldn't hinder progress. Please tell me, which approach do you prefer: freezing or deprecating?
Before tackling such rather cosmetic changes, I would like to get a thumbs-up for the general approach.

> `Could not load controller 'franka_gripper' because controller type 'GripperCommand' does not exist.`

The culprit is the definition of the [`franka_gripper` controller in `panda_moveit_config`](https://github.com/ros-planning/panda_moveit_config/blob/c38a59af4ca94c008b3677e1ee1787a64bc6fea3/config/simple_moveit_controllers.yaml#L14-L19). As I don't (yet) have access to a real robot, I wasn't able to test the real-robot control. However, I was assuming that you provide a [`GripperCommand` action](http://docs.ros.org/en/api/control_msgs/html/msg/GripperCommand.html) as well, don't you?

rickstaa:
> 



> > * Please revert the `prefix` to `arm_id` to avoid breaking changes.
> >   I get your point, that here it is used as a prefix to joint names, but in all our nomenclature we use the term `arm_id`. Especially for multi-arm situation it makes sense to distinguish multiple robots by ids.
> > * Please revert the renaming the `panda_arm.xacro` file for backwards compatibility
> 
> Note that we can maintain backwards compatibility by other means. For example, #210 illustrates how to deprecate a property in favour of a more meaningful name. A similar approach could be used for renaming the file. I fully agree with your request to maintain backwards compatibility. However, this shouldn't hinder progress. Please tell me, which approach do you prefer: freezing or deprecating? Before tackling such rather cosmetic changes, I would like to get a thumbs-up for the general approach.
> 
> > `Could not load controller 'franka_gripper' because controller type 'GripperCommand' does not exist.`
> 
> The culprit is the definition of the [`franka_gripper` controller in `panda_moveit_config`](https://github.com/ros-planning/panda_moveit_config/blob/c38a59af4ca94c008b3677e1ee1787a64bc6fea3/config/simple_moveit_controllers.yaml#L14-L19). As I don't (yet) have access to a real robot, I wasn't able to test the real-robot control. However, I was assuming that you provide a [`GripperCommand` action](http://docs.ros.org/en/api/control_msgs/html/msg/GripperCommand.html) as well, don't you?

I currently do not have access to the real robot, it is fully booked, but when I tested this last month, it worked. The controller definition looks good and should work as long as the [controller_msgs](http://wiki.ros.org/control_msgs) package is installed. Based on the real robot gripper code:

https://github.com/frankaemika/franka_ros/blob/e5a4bf16dfac61b9651bb18e820a6ed0a02cc929/franka_gripper/src/franka_gripper.cpp#L34

I do see why it would throw that error 🤔 . 

**EDIT:** Most of the times when I received that message, it was looking for the controller or [controller_manager](http://wiki.ros.org/controller_manager) on the old namespace (i.e. `panda`).

**EDIT2:** You can also check one of the packages I used with the robot. It should contain the right commits for the MoveIt integration to work https://github.com/rickstaa/panda-gazebo. Otherwise, I can check quickly investigate this problem in 3-4 weeks.

gollth:
> Note that we can maintain backwards compatibility by other means. For example, #210 illustrates how to deprecate a property in favour of a more meaningful name.
> A similar approach could be used for renaming the file. I fully agree with your request to maintain backwards compatibility.
> However, this shouldn't hinder progress. Please tell me, which approach do you prefer: freezing or deprecating?
> Before tackling such rather cosmetic changes, I would like to get a thumbs-up for the general approach.

Lets discuss this in #210 then. But, please, for this PR revert the changes which have nothing to do with detailed and coarse collision models

gollth:
> When I run `roslaunch panda_moveit_config franka_control.launch` on the latest `noetic-devel` branch of panda moveit config my controller manager is not able to start/find the `franka_gripper` controller:

So as it turnes out, the problem is not related to the definition of the franka_gripper controller in panda_moveit_config but rather, that the `ros_controllers.launch` file is trying to *start* the gripper controller, which doesn't exist. I proposed a fix [here](https://github.com/ros-planning/panda_moveit_config/pull/99)

gollth:
Hey @rickstaa + @rhaschke 

I wanted to test the new self-collision avoidance with the `*_sc` tags and their corresponding self-collision matrix in panda_moveit_config. A quick test was a bit disappointing, though. I was testing a cartesian PTP trajectory with the PILZ motion planner from the poses "rear" -> "front":

```xml
<group_state name="front" group="${name}">
  <joint name="panda_joint1" value="-0.9487613379359929" />
  <joint name="panda_joint2" value="-0.21695452212032515" />
  <joint name="panda_joint3" value="0.465433893620256" />
  <joint name="panda_joint4" value="-2.372735109191908" />
  <joint name="panda_joint5" value="0.11643626373543897" />
  <joint name="panda_joint6" value="2.1749505506913143" />
  <joint name="panda_joint7" value="0.22677080288864251" />
</group_state>

<group_state name="rear" group="${name}">
  <joint name="panda_joint1" value="-1.261635546123841" />
  <joint name="panda_joint2" value="-1.275479873955422" />
  <joint name="panda_joint3" value="-1.175981409376109" />
  <joint name="panda_joint4" value="-2.574291443811755" />
  <joint name="panda_joint5" value="-1.0959389218694833" />
  <joint name="panda_joint6" value="1.7193886261751965" />
  <joint name="panda_joint7" value="-1.1110643820436745" />
</group_state>
```

... which MoveIT happily plans:

moveit-close-plan

However executing this motion fails during the execution (as expected) since the internal control of the Panda will use the primitive collision capsules. This is also the place where the "collision" happens:

moveit-close-plan-fail

[According to the Self Collision matrix](https://github.com/ros-planning/panda_moveit_config/blob/kinetic-devel/config/panda_arm.xacro#L35-L53) `panda_link8_sc` + `panda_link{1,2}_sc` are not specifically disabled from the self collision checking. Why did MoveIT not recognize this then?

## Versions

* franka_ros -> this branch
* panda_moveit_config -> [fix/gripper](https://github.com/ros-planning/panda_moveit_config/pull/99) i.e [8d44c00](https://github.com/ros-planning/panda_moveit_config/pull/99/commits/8d44c0014bf22d0135ddd803f7d41df4b407369e) (plus the changes above with the two extra poses)
* MoveIt: noetic-devel i.e. `1.1.8`

franka_description/robots/utils.xacro
```diff
@@ -1,6 +1,62 @@
...
 <robot xmlns:xacro="http://www.ros.org/wiki/xacro">
+  <!-- ============================================================ -->
+  <!-- Macro to add an <interial> tag based on yaml-load properties -->
+  <!--                                                              -->
+  <!-- name:     Name of the robot link (without prefix)            -->
+  <!-- inertial: Dictionary of inertia properties                   -->
+  <!-- ============================================================ -->
+  <xacro:macro name="inertial" params="inertial:=^ name:=^">
+    <xacro:if value="$(arg gazebo)">
+      <xacro:unless value="${inertial.get(name, None)}">${xacro.warning('No inertia properties defined for: ' + name)}</xacro:unless>
```
gollth: The [recommended way](https://docs.python.org/3/reference/expressions.html#membership-test-operations) to test if a key exists in a dict would be: `<xacro:unless value="${name in inertial}">`

franka_description/robots/inertial.yaml
gollth: I debugged the problem for you and I think the problem is that your `link` and `inertial` macros are named similarly to _real_ URDF tags. In melodic the `xacro:...` prefix is not yet mandatory and thus xacro will try to interpret `<link>` as `<xacro:link>`... I suggest renaming these macros to be unique within URDF

gollth: Also `apt-get upgrade` is considered bad-practice in Dockerfiles AFAIR, because it would destroy reproducibility of docker images

franka_description/robots/utils.xacro
rhaschke: Fixed.

franka_description/robots/inertial.yaml
rhaschke: > I debugged the problem _for you_.

Thanks for looking into this. Your writing suggests that I am responsible for debugging this. However, I thought this is a collaborative effort from all sides :wink: In any case, I will adapt the macro names to be compatible with Melodic (which wasn't a primary goal when developing this PR).

rhaschke:
> > I fully agree with your request to maintain backwards compatibility. Please tell me, which approach do you prefer: freezing or deprecating? Before tackling such rather cosmetic changes, I would like to get a thumbs-up for the general approach.
> 
> Let's discuss this in #210 then. But, please, for this PR revert the changes which have nothing to do with detailed and coarse collision models

@gollth, I took the effort of removing all my cleanup commits from this PR. Nevertheless, please, comment on the originally proposed cleanup ideas and decide whether the (in #210) proposed mechanisms to maintain backwards compatibility are acceptable for you.

> [However executing this motion (from `rear` to `front`) fails during the execution since the internal control of the Panda will use the primitive collision capsules.](https://github.com/frankaemika/franka_ros/pull/199#issuecomment-1031591781)

Actually, collision checking between `hand_sc` and `link*_sc` was disabled. Collision checking was performed only for hand_sc and `link[0123]` (without `_sc`). This is fixed in the [latest commit](https://github.com/ros-planning/panda_moveit_config/commit/3821c4f0e551a86d2c88422ac759e272808ae9f2), making your example fail as expected.

Please also have a look at (and comment on) the TODOs I mentioned in the [OP](https://github.com/frankaemika/franka_ros/pull/199#issue-1048925821).

rhaschke:
CI sim failures:
```
/ros/src/franka_ros/franka_gazebo/test/franka_gripper_sim_test_with_object.cpp:82
The difference between duration and expected_duration is 0.15600000000000003, which exceeds allowed_absolute_duration_error, where
duration evaluates to 0.63600000000000001,
expected_duration evaluates to 0.47999999999999998, and
allowed_absolute_duration_error evaluates to 0.10000000000000001.
```
probably originate from now unified joint efforts. Was there a particular reason that Gazebo and real-world URDFs used different limits? For example, `joint1`, real:
```xml
<limit effort="87" lower="-2.8973" upper="2.8973" velocity="2.1750"/>
```
vs. Gazebo:
```xml
<limit effort="87" lower="-2.9671" upper="2.9671" velocity="2.5" acceleration="15"/>
```

gollth (approved):

gollth:
Hi @rhaschke 

sorry that this took so long. We had a bit of internal discussions about this PR, but finally found consensus. Thanks for the [fix in panda_moveit_config](https://github.com/ros-planning/panda_moveit_config/commit/3821c4f0e551a86d2c88422ac759e272808ae9f2) with the self-collision checks. Now this works as expected:

SRR-1202


I took the freedom to rebase your `modular-xacro` branch to our latest `develop` (which also contains some hotfixes for the failing CI tests). Also there is now a capsule macro like you suggested in your description.

Regarding the weird capsule at the Pilot: this was actually wrongly placed and rotated -> should also be fixed now. In addition this capsule is now part of link7 like you suggested.

franka_description/robots/inertial.yaml
gollth: I leave how you interpret and format my comments up to you. In my working environment I, as a PR author, feel responsible for my pull requests and am grateful for every help and feedback I get on them while not relying on others to fix the issues with it. But maybe this is different in the FOSS community 🤷🏻‍♂️

rhaschke:
Thanks for addressing the remaining TODOs. Looks good for me now as well. I just removed `link8_sc`, which became empty (doesn't have any visual or collision models associated).

> We had a bit of internal discussions about this PR, but finally found consensus.

I'm curious what was controversial about this PR.

gollth:
> I'm curious what was controversial about this PR.

Mostly naming and if the `_sc` links should be configurable via the `gazebo` XACRO arg or not. For now it should be fine, with the according documentation change (will push after this is merged)

PR #200, opened 2021-11-10
Define a JointTrajectoryController for franka_gazebo

rhaschke:
Replacement for #186, at least defining a joint trajectory controller at all.


gollth:
Thanks @rhaschke for this, but we were working on this already internally. See https://github.com/frankaemika/franka_ros/commit/e899240365797de6d53c222d6512b74c3a40268e for details. This also adds an `effort_joint_trajectory_controller` to `franka_control` with the gains from @rickstaa in https://github.com/frankaemika/franka_ros/pull/186#issuecomment-956144658.

PR #202, opened 2021-11-18
Only load descriptions if desired

fjulian:
Give the option to use a custom robot URDF when launch `franka_control` or `franka_gazebo`.


gollth:
Hi @fjulian and sorry for the late reply. Thanks for your interest in and your contribution to `franka_ros`. May I ask what use case you are covering with _not_ loading the URDF onto the parameter server?

fjulian:
Hi @gollth, thanks for the reply. The use case for this is to load a custom robot description, e.g. with a camera or a mobile base added to the arm. In that case, I would like to load a description in an outer launch file, but with the current launch files, it would be overwritten again. Having this flag would allow to avoid this.
Admittedly, this could also be solved with namespacing. However, I think it would be nice to have both options, allowing users to choose what they prefer, especially since my proposed solution keeps the default behavior the same.

rickstaa (approved):

PR #211, opened 2021-12-10
feat(franka_gazebo): add physics engine launch argument

rickstaa:
A simple pull request gives users the ability to change the physics engine. I used DART as the default physics engine since it appears to be the most stable (https://github.com/frankaemika/franka_ros/issues/160#issuecomment-989173918). The only downside of using this engine compared to the ODE engine is that the `get_physics_properties` service call does not yet support the physics engine [bullet].

Please note that the BULLET and SIMBODY physics engines do not work with the current model (see https://github.com/frankaemika/franka_ros/issues/160#issuecomment-989173918). dcc5eb0476d05300b6a2b02d407581f3f1f2a8f9 adds a warning to communicate this to the user.

### DIFF
--- a/franka_gazebo/launch/panda.launch
+++ b/franka_gazebo/launch/panda.launch
@@ -6,6 +6,7 @@
   <arg name="paused"      default="false" doc="Should the simulation directly be stopped at 0s?" />
   <arg name="world"       default="worlds/empty.world" doc="Filename to a SDF World for gazebo to use" />
   <arg name="rviz"        default="false" doc="Should RVIz be launched?" />
+  <arg name="physics"     default="ode"   doc="The physics engine used by gazebo"/> <!--Phyics engines: ode|dart-->
 
   <!-- Robot Customization -->
   <arg name="arm_id"      default="panda" doc="Name of the robot to spawn" />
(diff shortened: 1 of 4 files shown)


franka_gazebo/launch/panda.launch
```diff
@@ -7,6 +7,7 @@
   <arg name="paused"      default="false" doc="Should the simulation directly be stopped at 0s?" />
   <arg name="world"       default="worlds/empty.world" doc="Filename to a SDF World for gazebo to use" />
   <arg name="rviz"        default="false" doc="Should RVIz be launched?" />
+  <arg name="physics"     default="dart"   doc="The physics engine used by gazebo"/> <!--Phyics engines: dart|ode-->
```
rickstaa: @gollth, @marcbone I still think having this option is helpful since the dart simulation is more stable for me. We can, of course, also set ODE as the default since some Gazebo services are not yet implemented for the DART simulator.

rickstaa:
@gollth I just set the default physics engine to ODE (see  [6f45ba5](https://github.com/frankaemika/franka_ros/pull/211/commits/6f45ba5c87e0d0759a2229df04b893230c23b728)). I think this makes sense since dart is only preferred when doing effort/velocity control.

rickstaa:
@Maverobot, @gollth I rebased this pull request on the `develop` branch and solved the conflicts. Are you interested in merging this feature into the upstream, or can we close this PR 🤔?

rickstaa:
This will fix #362.

Issue #168, opened 2021-09-06
connected_to option in panda_gazebo.xacro


LucasJoseph:
It would be nice to have the `connected_to` argument in the panda_gazebo.xacro, similarly to panda_arm.xacro.

gollth:
Sure, we can do that. May I ask how do you use this arg? Do you include the `panda_arm.xacro` in another more complex URDF?

LucasJoseph:
Yeah, my robots are fixed on a table at different height and I include several `panda_arm.xacro` in my URDF.

gollth:
Alternatively you can also connect the robot in the Xacro where you instantiate the `panda_arm.xacro` macro, right? Something like:

```xml
<robot xmlns:xacro="http://www.ros.org/wiki/xacro" name="myrobot">
  <xacro:include filename="$(find franka_description)/robots/panda_gazebo.xacro" />

  <xacro:panda_arm arm_id="panda" />

  <link name="YOUR_TABLE_LINK" />

  <joint name="table_joint" type="fixed">
    <origin xyz="0 0 TABLE_HEIGHT" rpy="0 0 0" />
    <parent link="YOUR_TABLE_LINK" />
    <child  link="panda_link0" />
  </joint>

<robot>
```

Would that also work for you?

LucasJoseph:
Well, I could and that's what I did for now, but since the option is available in the `panda_arm.xacro` I wanted to be able to use it also with `panda_gazebo.xacro`

gollth:
Make sense to have the same interface in both files. Added in bd2bfd8. Thanks for the input

PR #212, opened 2021-12-14
Adds optional joint to connect the panda in panda_gazebo.xacro, similar to panda_arm.xacro

c-salmi:
Hi,

We're currently working on integrating the panda arm on a mobile base, and I noticed there is a difference between panda_gazebo.xacro and panda_arm.xacro. The parameter `connected_to` is missing which is required to connect the panda in gazebo to another link.

edit: after looking at the issues I saw #168 is exactly this


gollth:
Hi @c-salmi sorry I noticed your PR only when we already merged this feature internally. Is now done on latest `develop` (bd2bfd8). Thanks for your contribution

PR #213, opened 2021-12-14
feat(franka_gazebo): add 'log_level' parameter to 'FrankaGripperSim' gazebo plugin

rickstaa:
This pull request adds the 'franka_gripper/log_level' parameter to the 'FrankaGripperSim' Gazebo Plugin. This parameter allows users to get rid of the excessive logging of received action server commands. I need this in my fork since the logging slows down my training script. Another solution would be to change the `ROS_INFO_STREAM_NAMED` console logs to `ROS_DEBUG_STREAM_NAMED` (see #214).

### DIFF
--- a/franka_gazebo/include/franka_gazebo/franka_gripper_sim.h
+++ b/franka_gazebo/include/franka_gazebo/franka_gripper_sim.h
@@ -16,6 +16,7 @@
 #include <hardware_interface/joint_command_interface.h>
 #include <hardware_interface/robot_hw.h>
 #include <realtime_tools/realtime_publisher.h>
+#include <ros/console.h>
 #include <ros/time.h>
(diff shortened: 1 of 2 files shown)


gollth:
Which "excessive logging" actions are we talking about? Usually you should only get one line saying `New XYZ Action goal received ...`

gollth:
You can also change the verbosity on the fly with something like:

```
rosservice call /gazebo/set_logger_level "{ logger: 'ros.franka_gazebo.FrankaGripperSim', level: 'warn' }"
```

Can this help your performance problem?

rickstaa:
@gollth Thanks for your answer that solves my problem. I will close this pull request and #214.

gollth:
Glad I could help

PR #181, opened 2021-10-20
Adds position and velocity motion generators to FrankaHWSim

rickstaa:
This pull request adds the `hardware_interface/PositionJointInterface'`and `hardware_interface/VelocityJointInterface` interfaces to the `FrankaHWSim` module. When a user chooses these interfaces, motion generators (i.e. PID controllers) are used for controlling the robot.

- This pull request requires #157 to be merged first.
- I did not extensively tune the PID gains of the motion generators, but they seem to work. Feel free to adjust them further.
- There is one minor bug in the current implementation that you might want to solve in future versions. The arm moves slightly at the beginning of the simulation when the position and velocity motion generators are used. This is probably because the joint commands are not initialized to the starting joint angles but to `0`. As I'm a bit short on time, I did not investigate this issue further. #181 is, in my opinion, good enough for releasing the Noetic version of the panda_moveit_config.

### TODOS

- [ ] Tune PID gains of the motion generators if you want smoother behaviour.
- [ ] Adjust the  trajectory controller goal tolerance  (see https://github.com/frankaemika/franka_ros/issues/201).

### Other TODOS
- [ ] Update documentation.

PR #157, opened 2021-08-23
Adds the 'transmission' arg to the 'panda_arm.urdf.xacro' file

rickstaa:
This commit gives users the ability to set the hardware_interface that is used in the panda arm joints when loading the `panda_arm.urdf.xacro` file.

PR #224, opened 2022-02-15
Add simulation motion generators

rickstaa:
This commit combines #181 and #157 since they are closely related. I tested this pull request against the steps I explained in https://github.com/frankaemika/franka_ros/pull/181#issuecomment-1019301350. The position controllers seem to work, although as described in #181, the PID gains for the motion generators are not extensively tuned and could therefore be improved.


rickstaa:
@gollth Both #181 and #157 pass. Therefore, I think the CI pipeline failures here are due to timing issues and not related to the changes that this pr includes.

PR #227, opened 2022-02-22
Fix gazebo joint collision check bug

rickstaa:
This pull request ensures the gravity contribution is accounted for in the joint collision checking. This needs to be done since it is not stored in `joint->command` variable. Please have a look at #363 for more information.

### DIFF
--- a/franka_gazebo/src/joint.cpp
+++ b/franka_gazebo/src/joint.cpp
@@ -114,11 +114,11 @@ double Joint::getLinkMass() const {
 }
 
 bool Joint::isInCollision() const {
-  return std::abs(this->effort - this->command) > this->collision_threshold;
+  return std::abs(this->effort - this->command + this->gravity) > this->collision_threshold;
 }
 
 bool Joint::isInContact() const {
-  return std::abs(this->effort - this->command) > this->contact_threshold;
+  return std::abs(this->effort - this->command + this->gravity) > this->contact_threshold;
 }
 
 void Joint::setJointPosition(const double joint_position) {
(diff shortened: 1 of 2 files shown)


franka_gazebo/src/joint.cpp
```diff
@@ -64,10 +64,10 @@ double Joint::getLinkMass() const {
 }
 
 bool Joint::isInCollision() const {
```
rickstaa, line 116: @gollth, @marcbone Just a minor point, maybe you want to put a note in the documentation about how the simulated `cartesian_collision` and `joint_collision` differ from the real robot. I noticed that in the simulation, users do not need to send a reset command for the `<>_contact` variables to return to zero.

https://github.com/frankaemika/libfranka/blob/f1f46fb008a37eb0d1dba00c971ff7e5a7bfbfd3/include/franka/robot_state.h#L258-L274

https://github.com/frankaemika/libfranka/blob/f1f46fb008a37eb0d1dba00c971ff7e5a7bfbfd3/include/franka/robot_state.h#L258-L274

rickstaa: This comment was added to #363.

rickstaa:
@Maverobot, @gollth, since I still think this pull request fixes a critical bug, I just rebased it on the `develop` branch and ensured no CI errors were present. Would it be possible to merge this so I can use the upstream instead of my fork?

Your assistance in getting this pull request merged would be significantly appreciated 🙏🏻.

Thanks a lot in advance!

gollth (approved):
Thanks for finding this. Makes sense to me!

PR #226, opened 2022-02-16
feat(franka_gazebo): implement `set_franka_model_configuration` service

rickstaa:
~This PR implements a `reset_joint_states` service. This service can reset the joint states when pushed outside the joint limits. This can, for example, happen when Gazebo's `set_model_configuration` is used (see https://github.com/frankaemika/franka_ros/issues/225 for more information).~

#### Update

This PR implements a `set_franka_model_configuration`, which can be used to set call Gazebo's `set_model_configuration` service while ensuring that the reported joint positions stay within joint limits (see https://github.com/frankaemika/franka_ros/issues/225 for more information)

PR #228, opened 2022-02-22
fix(franka_gazebo): fix `tau_ext` internal controller bug

rickstaa:
This PR ensures that internal controller forces are not used in the external torque calculations when a joint is within its
joint limits. This needs to be done since, in that case, the joint boundaries already compensate for external torques (see https://github.com/frankaemika/franka_ros/issues/364). Feel free to change this pull request however you like, as there are multiple ways to solve this problem.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ Requires `libfranka` >= 0.8.0
   * `franka_description`: `<xacro:franka_robot/>` macro now supports to customize the `parent` frame and its `xyz` + `rpy` offset
   * `franka_hw`: Fix the bug where the previous controller is still running after switching the controller. ([#326](https://github.com/frankaemika/franka_ros/issues/326))
   * `franka_gazebo`: Add `set_franka_model_configuration` service.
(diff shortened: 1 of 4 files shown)


rickstaa:
@gollth, @marcbone Similar to my previous command https://github.com/frankaemika/franka_ros/pull/226#issuecomment-1044338514 it looks like something goes wrong in your clang checking action as it does not print the error it encounters. My local clang version does not throw any errors.

rickstaa:
@gollth, @marcbone Again I think your clang-format action is broken.

rickstaa:
Merging this pull request will solve #364.

rickstaa:
@Gollth, @FE-EnricoSartori,

I've just provided an update to the associated issue (#364) with detailed steps to reproduce the bug. Please take a moment to review #288 and share your thoughts. I'd particularly appreciate your input on whether the clamped command should be considered in collision checking instead of the user command. I was unsure if you also wanted to account for internal joint interactions when checking collisions.

Your feedback would be invaluable. Thank you!

PR #253, opened 2022-05-04
Added a namespace getter to the roslaunch Node

FReinwald:
This fixes a bug where the robot could not find its cartesian_impedance_example_controller if launched from a namespace that is not the default one.

How to reproduce:
Ubuntu 18.04
ros melodic
franka_ros Version 0.9.0

Run this launchfile or something similar with a robot in a namespace:
````
<?xml version="1.0"?>
<launch>

  <include file="$(find gazebo_ros)/launch/empty_world.launch" >
    <!-- Start paused, simulation will be started, when Pandas were loaded -->
    <arg name="paused" value="true"/>
    <arg name="use_sim_time" value="true"/>
  </include>

  <group ns="namespace_testing">
    <include file="$(find franka_gazebo)/launch/panda.launch">
      <arg name="arm_id"     value="panda_1" />
      <arg name="controller" value="cartesian_impedance_example_controller" />
      <arg name="rviz"       value="false" />
      <arg name="gazebo"     value="false" />
      <arg name="paused"     value="false" />
    </include>
  </group>

</launch>
````

You will observe that the robot will sway for some time as the controller does not work. After roughly 25 seconds you will get the following warning:

> [WARN] [1651669475.566877, 27.666000]: Controller Spawner couldn't find the expected controller_manager ROS interface.

I did not test this with other controllers or multiple robots but I assume they also have this problem.

### DIFF
--- a/franka_gazebo/scripts/delayed_controller_spawner.py
+++ b/franka_gazebo/scripts/delayed_controller_spawner.py
@@ -27,7 +27,7 @@ def main():
             break
 
     rospy.loginfo('Waking up. Spawning controller...')
-    node = roslaunch.core.Node(package, node_type, name=node_type, args=args)
+    node = roslaunch.core.Node(package, node_type, name=node_type, namespace=rospy.get_namespace(), args=args)
     launch = roslaunch.scriptapi.ROSLaunch()
     launch.start()
     launch.launch(node)


gollth:
Hi @FReinwald and thanks for your contribution. The bug you caught definitely needed addressing. However in the mean time the `delayed_controller_spawner` script was removed in favor for the `--wait-for TOPIC` option of the controlle spawner. See [here](https://github.com/frankaemika/franka_ros/commit/295b3ee7b845f9cb82e275b3169de492905eb7aa#diff-aa8d6dfa4aeca885a415e9f7208b8a7ced3f6141606b433cb95d2e1849c9655a) for details. 

On latest `develop` the launch file you posted spawns the controller as it should, therefore I would close this PR, but feel free to reopen if the issue is not solved in your opinion

PR #263, opened 2022-06-15
Fix(franka_example_controllers): Fix dynamic reconfigure node name for cartesian_impedance_example_con…

Olimoyo:
Add a missing / to the node name so that the node name is:

`/cartesian_impedance_example_controller/dynamic_reconfigure_compliance_param_node/`

instead of:

`/cartesian_impedance_example_controllerdynamic_reconfigure_compliance_param_node/`


gollth:
HI @Olimoyo 

thanks for the fix, absolutely makes sense. I added a changelog entry. As soon as the CI is through, we can merge this

PR #270, opened 2022-07-20
Add dual arm impedance controller for different target

pazeshun:
This PR adds DualArmCartesianPoseController to franka_example_controllers.
This controller is similar to [DualArmCartesianImpedanceExampleController](https://github.com/frankaemika/franka_ros/blob/0.9.0/franka_example_controllers/src/dual_arm_cartesian_impedance_example_controller.cpp) but is different from it in that this controller gets a different target pose (`geometry_msgs/PoseStamped`) for each arm.
This controller is useful for some teleoperations, where one haptic device (e.g., [Touch](https://www.3dsystems.com/haptics-devices/touch)) controls one arm.

This controller was originally written by @ykawamura96 ([original branch](https://github.com/pazeshun/franka_ros/tree/add_dual_arm_impedance_controller)) and refactored by @pazeshun

### DIFF
--- a/franka_example_controllers/CMakeLists.txt
+++ b/franka_example_controllers/CMakeLists.txt
@@ -33,9 +33,13 @@ endif()
 
 add_message_files(FILES
   JointTorqueComparison.msg
+  ArmsTargetPose.msg
 )
 
-generate_messages()
+generate_messages(DEPENDENCIES
+  std_msgs
+  geometry_msgs
+)
 
 generate_dynamic_reconfigure_options(
   cfg/compliance_param.cfg
@@ -79,6 +83,7 @@ add_library(franka_example_controllers
   src/cartesian_impedance_example_controller.cpp
   src/force_example_controller.cpp
   src/dual_arm_cartesian_impedance_example_controller.cpp
+  src/dual_arm_cartesian_pose_controller.cpp
   src/teleop_joint_pd_example_controller.cpp
   src/joint_wall.cpp
 )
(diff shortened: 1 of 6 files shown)


gollth:
Hi there

thanks for your interest in and contribution to `franka_ros`.

Could you describe a bit more the use case when one would need your `DualArmCartesianPoseController`? Glancing over the code it seems to me that it is combining two independent `CartesianImpedanceExampleController`s into one class. Couldn't the same be achieved with something like this:

```xml
<?xml version="1.0"?>
<launch>

  <arg name="robot_arm_id_right" value="right" />
  <arg name="robot_ip_right" />

  <arg name="robot_arm_id_left" value="left" />
  <arg name="robot_ip_right" />

  <group ns="$(arg robot_arm_id_right)">
    <include file="$(find franka_example_controllers)/launch/cartesian_impedance_example_controller">
      <arg name="robot_ip" value="$(arg robot_ip_right)" />
    </include>
  </group>

  <group ns="$(arg robot_arm_id_left)">
    <include file="$(find franka_example_controllers)/launch/cartesian_impedance_example_controller">
      <arg name="robot_ip" value="$(arg robot_ip_left)" />
    </include>
  </group>

</launch>
```
?

pazeshun:
@gollth Sorry for super late. Thank you for your advice.
Now we are testing two `CartesianImpedanceExampleController`s intead of `DualArmCartesianPoseController`.
Maybe it works in our application, so I'll close this PR for now.
If I find something bad about `CartesianImpedanceExampleController`, I'll reopen this or create another PR.
Thanks again!

gollth:
Sure no worries. All the best

PR #278, opened 2022-08-23
Fix move_to_start.launch

rhaschke:
To become compatible to noetic-devel branch of panda_moveit_config:
- Use `load_gripper` argument for both, `franka_control` and `move_group`
- Start controllers required by MoveIt


marcbone:
We refactored the move_to_start to not depend on MoveIt anymore. Therefore, I am going to close this.

PR #279, opened 2022-08-24
auto-recover from error

### LINKED ISSUE #232: franka_control_node incorrectly handles controller starting
1. Trying to start a `ros_control` controller when the robot is not in automatic mode (external activation device pushed down), `franka_control_node` correctly rejects the operation: 
`[ERROR] [ros.franka_control]: libfranka: Move command rejected: command not possible in the current mode!`
But: the controller is reported as started, e.g. in rqt's controller manager plugin.
2. If the robot is switched back to automatic mode (external activation device pulled up), starting a controller is still not possible:
    - `franka_control_node` doesn't report an error (anymore)
    - the controller is reported to be active in rqt, but the controller isn't actually running!
3. To enable starting a controller, one needs to re-launch `franka_control_node`, which is rather annoying if it is part of a larger roslaunch file.

rhaschke:
This implements auto-recovering from errors when the user-stop button is pressed+unpressed and thus the robot's mode switches to idle. Without that, one has to trigger the `error_recovery` action.
Partially fixes #232

### DIFF
--- a/franka_control/src/franka_control_node.cpp
+++ b/franka_control/src/franka_control_node.cpp
@@ -139,6 +139,7 @@ int main(int argc, char** argv) {
       }
     }
 
+    ROS_INFO_THROTTLE(1, "franka_control: controller activated");
     if (franka_control.connected()) {
       try {
         // Run control loop. Will exit if the controller is switched.
@@ -160,7 +161,7 @@ int main(int argc, char** argv) {
         has_error = true;
       }
     }
-    ROS_INFO_THROTTLE(1, "franka_control, main loop");
+    ROS_INFO_THROTTLE(1, "franka_control: main loop");
   }
 
   return 0;
--- a/franka_control/src/franka_control_node.cpp
+++ b/franka_control/src/franka_control_node.cpp
@@ -124,6 +124,10 @@ int main(int argc, char** argv) {
           control_manager.update(now, now - last_time);
           franka_control.checkJointLimits();
           last_time = now;
+
+          if (has_error && franka_control.robotMode() == franka::RobotMode::kIdle) {
+            has_error = false;
+          }
         } catch (const std::logic_error& e) {
         }
       } else {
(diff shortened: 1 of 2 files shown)


Ekanshh:
@rhaschke I faced this issue today, and it took me a lot of time to figure out this issue. Thanks for mentioning it and creating a PR. I hope they merge it asap.

franka_control/src/franka_control_node.cpp
```diff
@@ -124,6 +124,11 @@ int main(int argc, char** argv) {
           control_manager.update(now, now - last_time);
           franka_control.checkJointLimits();
           last_time = now;
+
+          if (has_error && franka_control.robotMode() == franka::RobotMode::kIdle) {
+            franka_control.robot().automaticErrorRecovery();
```
andrejpan: This line is unnecessary. We will remove it and merge this PR.

Please let us know if we are missing something.

Edit: This will be merged latest first week of 2023.

rhaschke: Which line exactly do you refer to? As you already force-pushed, I cannot easily see my suggested code anymore.
But in a different place, you call `automaticErrorRecovery()` as well:
https://github.com/frankaemika/franka_ros/blob/c392c9112e1184f6dad5437f4453cb8020de4219/franka_control/src/franka_control_node.cpp#L53-L54
Why is it unnecessary here?

Maverobot: We meant the line ` robot.automaticErrorRecovery(); `. In the `if` condition, being in `franka::RobotMode::kIdle` mode already means that no error is present. Therefore, the ` robot.automaticErrorRecovery();` is unnecessary. 

At that different place, if the mode is `franka::RobotMode::kIdle`, then `robot.automaticErrorRecovery();` would also be unnecessary. We just always call it there for simplicity.

@rhaschke By the way, do you think merging this PR is enough for closing the issue https://github.com/frankaemika/franka_ros/issues/232 ?

marcbone (approved):

Maverobot (approved):

PR #292, opened 2022-10-16
Fix various issues in TriggerRate

rhaschke:
In order to actually reach the desired rate, the inequality needs to be replaced with greater equal.
Here is an example with `publish_rate = 50` and Gazebo simulation (i.e. perfect sim time)

**Before**: rate = 47.619 Hz = 1/21ms
```
rostopic hz /franka_state_controller/joint_states
subscribed to [/franka_state_controller/joint_states]
WARNING: may be using simulated time
average rate: 47.619
        min: 0.019s max: 0.024s std dev: 0.00086s window: 143
```
**After**: rate = 50 Hz = 1/20ms (as expected)
```
average rate: 50.000
        min: 0.018s max: 0.022s std dev: 0.00063s window: 52
```

### DIFF
--- a/franka_hw/include/franka_hw/trigger_rate.h
+++ b/franka_hw/include/franka_hw/trigger_rate.h
@@ -13,7 +13,7 @@ class TriggerRate {
 
  private:
   ros::Time time_stamp_;
-  double period_;
+  ros::Duration period_;
 };
 
 };  // namespace franka_hw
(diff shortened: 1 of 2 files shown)


rhaschke:
I found some more issues with this class:
- Using ros::Time() conversion to double can introduce rounding errors, causing sporadic misses in state publishing
- When ros::Time was reset (e.g. in simulation), there were no state updates at all, because `now - time_stamp_` is negative.

rickstaa:
@rhaschke Excellent observations! Much appreciated! I applied them to my fork.

@Maverobot Unfortunately, I couldn't test this PR on the real robot, but they look good. It would be great if these could be merged as they will significantly improve the performance ⚡.

Issue #301, opened 2022-11-20
[franka_description] Should mention paper's friction model is not used?


EricCousineau-TRI:
In context of https://www.franka-community.de/t/panda-dynamic-model/4573/3

From brief inspection, it looks like the friction model from the paper (https://hal.inria.fr/hal-02265293/) is not used.
Consider mentioning in the docs?

May relate #254, #249

Issue #300, opened 2022-11-20
[franka_description] Docs say `gazebo=true` triggers use of paper's estimated paremters, but they are always used?


EricCousineau-TRI:
From docs:
https://github.com/frankaemika/docs/blame/12fb63aa406b3e33dfc57a97ad5a5362e1ad8aad/source/franka_ros.rst#L36-L39

However, upon inspection, it looks like the inertial parameters are *not* dependent on the `gazebo` parameters?
https://github.com/frankaemika/franka_ros/blob/a58d3052a241304392847df6464234c83d728a38/franka_description/robots/common/utils.xacro#L4-L6

Is this intended? If so, can the docs be updated?
If not, can this possibly be fixed?

Relates
https://github.com/frankaemika/franka_ros2/issues/11
https://www.franka-community.de/t/panda-dynamic-model/4573/3

gollth:
Hi @EricCousineau-TRI

sorry for the late reply. The inertia parameters _are_ dependent on the `gazebo` arg but not in the place you searched for. Rather the condition is placed a little bit further down in the `link_with_sc` macro:

https://github.com/frankaemika/franka_ros/blob/2d458abad7390bb73771c91787c361c2b5cfa6ad/franka_description/robots/common/utils.xacro#L51-L53

You can also verify that by simply inspecting the URDFs ;)

```console
$ xacro $(rospack find franka_description)/robots/fr3/fr3.urdf.xacro | grep "<inertial>" | wc -l
0

$ xacro $(rospack find franka_description)/robots/fr3/fr3.urdf.xacro gazebo:=true | grep "<inertial>" | wc -l
8
```

I hope this answers your question. Feel free to reopen the ticket if not.

All the best

PR #302, opened 2022-11-20
[franka_description] Use proper title; make more searchable

EricCousineau-TRI:
During investigation that yielded #300 #301

### DIFF
--- a/franka_description/robots/common/inertial.yaml
+++ b/franka_description/robots/common/inertial.yaml
@@ -1,10 +1,11 @@
-# This file does not contain official inertial properties
-# of panda robot. The values are from the identification
-# results published in: Identification of the Franka Emika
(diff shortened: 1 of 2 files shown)


rickstaa (approved):
@EricCousineau-TRI, thank you for submitting this pull request. I believe it's a valuable addition, as it provides essential information to users about the origin of simulated inertia values – great job! 👍🏻

@Maverobot, I apologize for tagging you in multiple PRs, but considering that this pull request only modifies docstrings, I believe it can be merged seamlessly without any concerns. Your input on this would be appreciated.

rickstaa:
@EricCousineau-TRI, would it be possible to consider changing the PR title and commit message to something like:

docs(franka_gazebo): add inertia properties reference

This adjustment would make it more evident at a glance what the PR is focused on, as the current title doesn't immediately convey its purpose.

PR #305, opened 2022-11-28
Add missing install of franka_gazebo's models and world folders

rhaschke:
Without this, the [Pick+Place example](https://frankaemika.github.io/docs/franka_ros.html#pick-place-example) doesn't work from install space.


simonbogh:
Would be nice to get this merged so that all needed files are available and the documentation examples work as intended.

rickstaa (approved):
Hey @gollth and @Maverobot,

I've reviewed this pull request, and it successfully addresses the issue reported in https://github.com/frankaemika/franka_ros/issues/360. Considering that the current state of the franka_gaebo binary includes broken examples, this fix is essential. It's preventing users from easily utilizing the examples, models and worlds your package provides.

Would it be possible to move forward and merge this pull request to ensure a smoother experience for the end users?

Issue #304, opened 2022-11-25
Gazebo model of the panda robot not responding to velocity inputs


sabeaussan:
When using a simple JointGroupVelocityController with gazebo, the robot does not respond to command but start slowly drifting from original position even when commands is set to zero. After checking the controller, commands are received and set in the "update" function but ignored by the model. However, things behave as expected with real hardware so not sure if the problem could come from the ros_control_gazebo plugin or the franka_hw_sim. Note that JointGroupPositionController works well.

JohannesPankert:
There is a bug in the gazebo plugin. This PR fixes it:
https://github.com/frankaemika/franka_ros/pull/306

sabeaussan:
Great ! Thank you. Closing this issue.

PR #306, opened 2022-11-28
fix gazebo velocity control

JohannesPankert:
The efforts computed by the velocity PID controllers were discarded before. Now they are applied to the joints just like with the position control PIDs.


JohannesPankert:
#304

PR #311, opened 2022-12-26
Add missing arm_id parameter to JointVelocityExampleController config

rhaschke:
With Gazebo, the `JointVelocityExampleController` fails with the following error:
`[ERROR] [1671708722.634915401] [ros.franka_example_controllers] [/mujoco_server]: JointVelocityExampleController: Could not get parameter arm_id`


rickstaa:
@rhaschke, good catch. Thanks! @Maverobot would it be possible to merge this minor fix since it prevents the [joint_velocity_example_controller](https://github.com/frankaemika/franka_ros/blob/develop/franka_example_controllers/launch/joint_velocity_example_controller.launch) from working.

rickstaa (approved):

PR #320, opened 2023-01-24
Fix order of include directories

### LINKED ISSUE #319: Build error: ’class franka_hw::FrankaHW’ has no member named ‘robotMode’
Since #279 was merged, I get the following compiler error when building `franka_ros` from source _and_ having its Debian packages installed in `/opt/ros` too:
```
franka_control_node.cpp:128:43: error: ‘class franka_hw::FrankaHW’ has no member named ‘robotMode’; did you mean ‘robotMutex’?
```

rhaschke:
The reason for #319 is that `libfranka_INCLUDE_DIRS` points to `/opt/ros/*/include` and thus also pulls in all `franka_ros` include directories _before_ those from the overlay. The only clean solution is to sort include directories by topological order.
Fixes #319.

### DIFF
--- a/franka_gripper/CMakeLists.txt
+++ b/franka_gripper/CMakeLists.txt
@@ -19,6 +19,10 @@ if(NOT Franka_FOUND)
   find_package(Franka 0.8.0 REQUIRED)
 endif()
 
+# merge Franka + catkin LIBRARIES/INCLUDE_DIRS in topological order
+list_insert_in_workspace_order(catkin_LIBRARIES ${Franka_LIBRARIES} ${catkin_LIBRARIES})
+list_insert_in_workspace_order(catkin_INCLUDE_DIRS ${Franka_INCLUDE_DIRS} ${catkin_INCLUDE_DIRS})
+
 add_action_files(
   DIRECTORY action
   FILES Grasp.action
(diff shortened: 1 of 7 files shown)


tropioneer (approved):

Maverobot (dismissed):
For more information about the issue: https://github.com/ros/catkin/issues/200

franka_gripper/CMakeLists.txt
```diff
@@ -19,6 +19,10 @@ if(NOT Franka_FOUND)
   find_package(Franka 0.8.0 REQUIRED)
 endif()
 
+# merge Franka + catkin LIBRARIES/INCLUDE_DIRS in topological order
+list_insert_in_workspace_order(catkin_LIBRARIES ${Franka_LIBRARIES} ${catkin_LIBRARIES})
```
Maverobot: 2 things:

* You merged `Franka_LIBRARIES` and `catkin_LIBRARIES` into `catkin_LIBRARIES` here but the `Franka_LIBRARIES` is not removed below. Is it intentional?
* Why only merge the `XXX_LIBRARIES` in `franka_gripper` package?

rhaschke: Very well-spotted! This is a left-over of my initial work, where I started to merge both, `INCLUD_DIRS` and `LIBRARIES`. However, it turned out, that it is sufficient to merge `INCLUDE_DIRS`. `LIBRARIES` shouldn't be touched.
Should be fixed now.

Maverobot (approved):

andrejpan (approved):

franka_gripper/CMakeLists.txt
Maverobot: Thanks a lot for the contribution! It looks really good now.

Issue #334, opened 2023-04-07
Reading Efforts and Velocities in Gripper node


ayralth:
Hello,

I wanted to read grasping forces of the gripper, but the franka_gripper_node publishes a state topic with gripper-joints efforts and velocities set to zero.

https://github.com/frankaemika/franka_ros/blob/41a3c7808b1a141a242a1258cbdbfebc76ae2a7a/franka_visualization/src/gripper_joint_state_publisher.cpp#L44

Can you clarify what is going on here ? Is there a way to read gripper efforts with franka_ros or libfranka ?
Thanks

marcbone:
Unfortunately, there is no way to read the gripper speed and the gripper efforts. Thats why there are always zero. On the other side when you use the gripper to grasp you always set the speed and grasping force. Best you can do is to calculate the speed yourself. If you use libfranka directly you can also get a boolean that states if you are currently grasping. The grasping force is then the one which you commanded.

gavanderhoorn:
Would it perhaps be more semantically correct to not set those entries to `0.0` (ie: not set them at all)?

Those lists containing entries basically implies whatever is in those entries *is* the actual state of those joints.

The way `gripper_joint_state_publisher` is implemented, there'd be no way to determine whether the state is really `0.0` (ie: no velocity, no effort), or something else.

The [documentation](http://docs.ros.org/en/melodic/api/sensor_msgs/html/msg/JointState.html) of `sensor_msgs/JointState` seems to suggest things are optional:

> This message consists of a multiple arrays, one for each part of the joint state.
> The goal is to make each of the fields optional. When e.g. your joints have no effort associated with them, you can leave the effort array empty.

rucha6198:
**Why is gripper velocity and effort always zero?**

I am trying to read the velocity of the gripper of Franka Emika Panda in the real-world. However, I always get the gripper velocity and effort to be zero. I get correct values for gripper velocity and effort when I use the Panda in Gazebo simulation.

I get correct values for other joints in Gazebo as well as for real-world Panda.

Could anyone please help me out? I don't understand what could be going wrong.
Thank you!

PR #335, opened 2023-04-11
Don't set vel or eff for gripper joints

gavanderhoorn:
Context: frankaemika/franka_ros#334.

See discussion there and / or the commit comment.

I've unfortunately not tested this, just wanted to provide a start.

### DIFF
--- a/franka_visualization/src/gripper_joint_state_publisher.cpp
+++ b/franka_visualization/src/gripper_joint_state_publisher.cpp
@@ -21,10 +21,8 @@ int main(int argc, char** argv) {
   ros::Rate rate(publish_rate);
 
   sensor_msgs::JointState states;
-  states.effort.resize(joint_names.size());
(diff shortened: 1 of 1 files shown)


marcbone:
Thx for the PR. However, this wont do much since this node is only used when you start the franka_visualization launch file (only few people do that). You also need to change it in the [gripper node](https://github.com/frankaemika/franka_ros/blob/develop/franka_gripper/src/franka_gripper_node.cpp). Further, somebody should check what the current behavior in gazebo is :sweat_smile:

gavanderhoorn:
I could add the changes to `franka_gripper`, but I won't have time to analyse the gazebo plugin.

Let me know whether you'd rather I close this PR.

gavanderhoorn:
Closing, as it's incomplete.

Perhaps it helps future contributors.

PR #359, opened 2023-08-24
refactor: improve python code formatting

rickstaa:
This pull request improves the Python code following the [black](https://github.com/psf/black) format.

### DIFF
--- a/franka_example_controllers/scripts/dual_arm_interactive_marker.py
+++ b/franka_example_controllers/scripts/dual_arm_interactive_marker.py
@@ -7,10 +7,11 @@
 import rospy
 import argparse
 
-from interactive_markers.interactive_marker_server import \
-    InteractiveMarkerServer, InteractiveMarkerFeedback
-from visualization_msgs.msg import InteractiveMarker, \
-    InteractiveMarkerControl, Marker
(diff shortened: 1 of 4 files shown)


andrejpan:
Thanks for the effort, however this project should first accept (automated) Python format rules (and so far has none). Besides this PR, nobody else knows that formatting should be according to black.

Maybe open another PR where you integrate automated Python formatting and lining rules.

rickstaa:
@andrejpan, thanks for the fast response. I just checked your CI, and it looks like you already use a CI step to check Python formatting with [pycodestyle](https://pypi.org/project/pycodestyle/). As pycodestyle is close but not fully equal to black I think we can close this PR.

https://github.com/frankaemika/franka_ros/blob/d439fc7d66a3edb890eefb0fa955f6d7042cac12/.github/workflows/ci.yml#L78-L87

We could, in the future, use a pre-commit hook, with [pre-commit](https://pre-commit.com/#pre-commit-configyaml---hooks), to automatically format the python code according to pycodestyle (see https://www.pydocstyle.org/en/4.0.0/usage.html#usage-with-the-pre-commit-git-hooks-framework).

PR #365, opened 2023-09-03
fix(franka_hw_sim): fix robot name differ warning

rickstaa:
This pull request ensures that the `Caution: Robot Names differ!` warning is not thrown when the user omits the `<robotNamespace>` tag in their URDF/SDF file.


rickstaa:
replaced by #366.

PR #370, opened 2023-09-12
Remove enforcement of C++14 standard

rhaschke:
Remove enforcement of C++14 standard. Modern defaults are:
- C++14 on Ubuntu 20.04 / gcc 9.4
- C++17 on Ubuntu 22.04 / gcc 11.4

Thus, C++14 is satisfied automatically on these systems.
However, on Ubuntu 22.04, log4cxx requires C++17 (which is the default).
Enforcing C++14 downgrades the standard, rendering log4cxx uncompilable.

### DIFF
--- a/franka_example_controllers/src/cartesian_impedance_example_controller.cpp
+++ b/franka_example_controllers/src/cartesian_impedance_example_controller.cpp
@@ -10,7 +10,8 @@
 #include <pluginlib/class_list_macros.h>
 #include <ros/ros.h>
 
-using namespace boost::placeholders;
+using boost::placeholders::_1;
+using boost::placeholders::_2;
 
 #include <franka_example_controllers/pseudo_inversion.h>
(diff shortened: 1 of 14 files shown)


franka_example_controllers/src/cartesian_impedance_example_controller.cpp
```diff
@@ -10,6 +10,9 @@
 #include <pluginlib/class_list_macros.h>
 #include <ros/ros.h>
 
+using boost::placeholders::_1;
```
rickstaa, line 13: @rhaschke, quick question. Why are these placeholders added? I don't see them being used 🤔.

rickstaa (approved):
Looks good to me! Added it to my fork. Thanks for his pull request.

v4hn:
still (and by now more) relevant.

PR #381, opened 2024-03-31
fix(franka_hw): include cstdint in franka_hw/resource_helpers

henrispriet:
`franka_hw` wouldn't build for me because it couldn't find type `uint8_t`, this fixes it.

cc: https://github.com/lopsided98/nix-ros-overlay/issues/382


lopsided98:
I think this should go in `resource_helpers.h` instead

PR #298, opened 2022-11-11
[franka_hw] Install plugin xml to enable to load FrankaCombinableHW from apt version of franka_hw

pazeshun:
When I try to load `franka_hw/FrankaCombinableHW` from apt version of `franka_hw`, I faced the following error:
```
[ERROR] [/dual_panda] [1668147868.535937528]: Skipped loading plugin with error: XML Document '/opt/ros/melodic/share/franka_hw/franka_combinable_hw_plugin.xml' has no Root Element. This likely means the XML is malformed or missing..
[ERROR] [/dual_panda] [1668147868.537817487]: Could not load robot HW 'rarm' because robot HW type 'franka_hw/FrankaCombinableHW' does not exist.
[ERROR] [/dual_panda] [1668147868.555089521]: franka_combined_control_node:: Initialization of FrankaCombinedHW failed!
```
This is because `franka_hw/franka_combinable_hw_plugin.xml` is not installed:
https://github.com/frankaemika/franka_ros/blob/a58d3052a241304392847df6464234c83d728a38/franka_hw/CMakeLists.txt#L104-L112

This PR installs `franka_hw/franka_combinable_hw_plugin.xml`.
I confirmed this PR works by:
```
catkin init
catkin config --install
catkin build
source <WORKSPACE_PATH>/install/setup.bash
```

cf. https://answers.ros.org/question/274849/gmapping-skipping-xml-document/

PR #394, opened 2024-06-20
Remove reference to non-existing plugin file

rhaschke:
`franka_combinable_hw_plugin.xml` doesn't exist. The reference in package.xml triggers the error:
`Skipped loading plugin with error: XML Document '/opt/ros/one/share/franka_hw/franka_combinable_hw_plugin.xml' has no Root Element. This likely means the XML is malformed or missing..`


pazeshun:
Could you check #298?
I think `franka_combinable_hw_plugin.xml` exists in source, but is not installed properly.

rhaschke:
Thanks for the hint. I didn't notice the existence of the file.

